PR #2, opened 2016-10-04
Trace system

nuft:
To have better debug info I need the possibility to trace events in software.

This is largely inspired by earlier work from @antoinealb - I just needed more data stored in the trace events (int, float, pointers and strings).

There is still an issue: For some reason the event timestamp is always 0. Has anyone experience with `chVTGetSystemTimeX()` and knows what is needed to get a useful value from it?

### DIFF
--- a/master-firmware/src/trace/trace.c
+++ b/master-firmware/src/trace/trace.c
@@ -2,16 +2,16 @@
 #include <stddef.h>
 #include <string.h>
 
-#ifdef _CHIBIOS_RT_
-#include <ch.h>
-#define TRACE_LOCK() syssts_t __syssst = chSysGetStatusAndLockX()
-#define TRACE_UNLOCK() chSysRestoreStatusX(__syssst)
-#define TRACE_TIMESTAMP_GET() ST2MS(chVTGetSystemTimeX())
-#else /* TEST */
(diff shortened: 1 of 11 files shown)


master-firmware/src/trace/trace.c
```diff
@@ -0,0 +1,145 @@
+#include "trace.h"
+#include <stddef.h>
+#include <string.h>
+
+#ifdef _CHIBIOS_RT_
```
antoinealb: Are you sure that condition is true ? This is defined in `ch.h`, not by a compiler define. I think your issue with timestamps might come from this.

SyrianSpock: I would consider an `#ifdef TESTS` instead

antoinealb: Or put it in a different translations unit and use functions instead of macros. I am not sure it would be significantly slower, especially if we start using LTO.

antoinealb:
Are the timestamp correct with the latest commits ?

nuft:
I need to test on hardware to be sure that I'm locking correctly, but the assembly output looks correct.

nuft:
I tested this on hardware: works fine.
If you're ok with this, please merge.

PR #4, opened 2016-10-20
Use protobuf on messagebus

antoinealb:
This implements publish/subscribe with serialized buffers using Protobuf. Currently the CI build will be broken by this, I need to figure out how to install the Protobuf compiler into Travis CI.

@SyrianSpock ok?

### DIFF
--- a/master-firmware/src/can/uavcan_node.cpp
+++ b/master-firmware/src/can/uavcan_node.cpp
@@ -13,7 +13,9 @@
 #include <cvra/Reboot.hpp>
 #include <cvra/StringID.hpp>
 #include <cvra/proximity_beacon/Signal.hpp>
+#include "proximity_beacon.pb.h"
 #include <msgbus/messagebus.h>
+#include <messagebus_protobufs.h>
 #include "robot_parameters.h"
 #include "motor_driver.h"
 #include "motor_driver_uavcan.h"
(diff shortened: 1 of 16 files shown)


SyrianSpock:
Tested and working on hardware. I'll wait until Travis is fixed to merge.

master-firmware/src/can/uavcan_node.cpp
```diff
@@ -232,7 +234,7 @@ void main(void *arg)
     static messagebus_topic_t proximity_beacon_topic;
     static MUTEX_DECL(proximity_beacon_topic_lock);
     static CONDVAR_DECL(proximity_beacon_topic_condvar);
-    static float proximity_beacon_topic_value[2];
+    static uint8_t proximity_beacon_topic_value[ProximityBeaconSignal_size];
```
antoinealb: The buffer in the protobuf macros is only used for encoding before publishing. You still need one static buffer to hold the topic data.

SyrianSpock: Oh I see, I didn't notice it was used int `messagebus_topic_init` because there was no diff, my bad.
But can we rename it for clarity ?

antoinealb:
Yes, I will rename it when I start working on that PR again.

PR #5, opened 2016-10-20
Fix obstacle avoidance

antoinealb:
This PR fixes a bug in the obstacle avoidance (path planner) module. It also removes the fast math library, which is not related to this bug but not very useful with recent MCUs.

It also adds some unit tests for the obstacle avoidance module.

@SyrianSpock ok?


SyrianSpock:
Looking good. I will test on hardware ASAP and merge.

SyrianSpock:
Works on hardware, I'll add integration code in a following PR.

PR #8, opened 2016-10-26
Trajectory manager tests

antoinealb:
This PR implements a few "smoke" tests for the trajectory manager. I did not want to test it completely, but at least we have some basic infrastructure to use when we want to change `trajectory_manager`.


SyrianSpock:
I think this is a great use of tests, it formalizes our understanding of how the library works.

PR #11, opened 2016-11-05
Import uavcan datatypes

nuft:
Define all UAVCAN datatypes locally.
As discussed [here](https://github.com/cvra/motor-control-uavcan-dsdl/issues/5).


SyrianSpock:
Shouldn't we just merge everything in one PR ?

nuft:
> Shouldn't we just merge everything in one PR ?

Ok, if you prefer.

PR #12, opened 2016-11-05
Import motor control source

nuft:
- import [cvra/motor-control-firmware](https://github.com/cvra/motor-control-firmware)
- define all UAVCAN datatypes locally. As discussed [here](https://github.com/cvra/motor-control-uavcan-dsdl/issues/5)


nuft:
Todo:
* integrate CI for motor control firmware
* verify correct migration with test on hardware
* cleanup

Also, what do you think of moving the lwip, ChbibiOS and uavcan submodules into `\lib`?

SyrianSpock:
Moving all 3rd party software to `lib` sounds good to me, this way we ensure one version is used on all boards.

nuft:
Another thing to note is that we lose the git history of the motor-control firmware.

SyrianSpock:
1. If we merge with subtree, we squash everything and lose the history in the global tree, but we keep the tree in the directory where the external repo was merged. This makes it easier if we need to backport changes to the original repo
2. If we merge the two repos to make it look like it was the same repo all along, we end up messing the history because everything gets mixed, so we lose the utility of keeping the history, and we lose the backport option
3. If we merge the external repo as if it was a feature branch, all of its commits are kept in a continuous chunk that comes after the history of the main repo at the time of merge. We lose the backport option and we keep the history, but that history is no good to use because we won't run a bisect on the 300+ commits of the motor-control-firmware to find a regression in the master-firmware

So using option 1. and losing the history of the `motor-control firmware` seems like the lesser evil to me.

PR #15, opened 2016-11-08
Dynamic obstacle avoidance

SyrianSpock:
This is PR implement obstacle avoidance behavior using the signal from the proximity beacon.
Currently it allows the robot to stop when an opponent is close (and in direction of motion of the robot), and retry.

To do in this PR:
- [x] Add opponent to the map and query a new path to destination
- [x] Fix dynamic avoidance issues
- [x] Pass user acceptance testing 

Need to be done some day
- Move robot out of obstacle (if any) before planning a new path.
- Support two opponents (only one supported for now)


master-firmware/src/can/uavcan_node.cpp
```diff
@@ -230,7 +232,7 @@ void main(void *arg)
     static messagebus_topic_t proximity_beacon_topic;
     static MUTEX_DECL(proximity_beacon_topic_lock);
     static CONDVAR_DECL(proximity_beacon_topic_condvar);
-    static float proximity_beacon_topic_value[2];
+    static float proximity_beacon_topic_value[3];
```
antoinealb: Maybe replace that with a struct ?

master-firmware/src/dynamic_obstacle_avoidance.c
```diff
@@ -0,0 +1,47 @@
...
+
+#define DYNAMIC_OBSTACLE_AVOIDANCE_STACKSIZE 128
+
+typedef struct {
+    float distance;
+    float angle;
+} beacon_signal_t;
+
+
+void dynamic_obstacle_avoidance_policy(void* _robot)
```
antoinealb: I am not too sure about having the dynamic obstacle avoidance policy in another thread from strategy is a wise choice, but I think it could be easy enough to merge in the `trajectory_test_end` function, like Microb Technology did, if we need to later on

master-firmware/src/strategy.c
```diff
@@ -55,14 +56,55 @@ void strategy_goto_avoid(
     NOTICE("Path to (%d, %d) computed with %d points\r\n", x_mm, y_mm, num_points);
 
     /* Execute path, one waypoint at a time */
-    for (int i = 0; i < num_points; i++) {
-        NOTICE("Going to x: %.1fmm y: %.1fmm\r\n", points[i].x, points[i].y);
-        trajectory_goto_xy_abs(robot_traj, points[i].x, points[i].y);
-        trajectory_wait_for_finish(robot_traj);
+    for (int j = 0; j < num_retries; j++) {
+        NOTICE("Try #%d", j);
```
antoinealb: maybe use `DEBUG` for those

master-firmware/src/robot_helpers/math_helpers.c
```diff
@@ -12,3 +15,24 @@ float angle_delta(float start, float end)
         return res_conj;
     }
 }
+
+bool math_point_is_in_square(poly_t* square, int x, int y)
+{
```
antoinealb: There is already `is_in_poly` in `polygon.h`, what is the reason for this function ? Or at least it could be a wrapper around `is_in_poly`.

master-firmware/src/robot_helpers/trajectory_helpers.c
```diff
@@ -1,54 +1,125 @@
...
+    if ((watched_end_reasons & TRAJ_END_COLLISION) && bd_get(&robot->distance_bd)) {
+        return TRAJ_END_COLLISION;
+    }
+
+    if ((watched_end_reasons & TRAJ_END_COLLISION) && bd_get(&robot->angle_bd)) {
+        return TRAJ_END_COLLISION;
+    }
+
+    if (watched_end_reasons & TRAJ_END_OPPONENT_NEAR) {
+        beacon_signal_t beacon_signal;
```
antoinealb: For now it only handles a single opponent, correct ?

master-firmware/src/robot_parameters.h
```diff
@@ -11,6 +11,7 @@ extern "C" {
  * Robot base parameters
  */
 #define ROBOT_SIZE_X_MM                     205 // Front to back length
+#define DEFAULT_OPPONENT_SIZE_MM            300
```
antoinealb: Ok for now, but maybe we will move it to `config.yml`.

master-firmware/src/robot_helpers/math_helpers.c
SyrianSpock: It was implemented to circumvent the bug encountered with `is_in_poly`. Now that the bug is fix I should go back to using `is_in_poly`

master-firmware/src/robot_helpers/trajectory_helpers.c
SyrianSpock: True, this is a design choice for now, we'll worry about support for two opponents later. I'll create a related issue for this.

master-firmware/src/robot_parameters.h
SyrianSpock: Yes, we need to unify all parameters and choose how it will support the robot specific parameters.

PR #18, opened 2016-11-12
Add DroneCI

antoinealb:
This pull request adds basic support for drone CI. So far it uses my custom server, but we will be able to migrate to CVRA's when we buy one.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,7 +1,11 @@
+clone:
+  recursive: true
+
 build:
   master-firmware-tests:
     image: cvra/cpputest
     commands:
+      - git submodule update --init --recursive
       - cd master-firmware
       - packager
       - mkdir build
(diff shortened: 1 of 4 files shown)


SyrianSpock:
Should probably also update the readme to show the build status of all the different firmwares/tests

antoinealb:
Thing is you only get one badge for the whole project, not one per sub build.

antoinealb:
But we could build our own badge server, e.g using https://github.com/gcmalloc/epfl-moodle

SyrianSpock:
Ok, but can we at least update the current badge with the results from Drone CI ?
We can add other badges later.

antoinealb:
Sure. Should I put my temporary server URL for the badge or we wait for the final infrastructure ?

PR #20, opened 2016-12-07
Add a tool to find all callers for a given function

antoinealb:
This can be useful, i.e. when tracking down why a binary is getting very large.

The callee name is given as a regular expression, which is useful when the
exact matching name is not known (e.g. free vs _free_r).

Example invocation used to track down why printf gets included in a build:

```
$ call_explorer.py can-io-firmware/build/ch.list malloc
fiprintf      ->  _vfiprintf_r
__assert_func ->  fiprintf
_vfiprintf_r  ->  __sbprintf
__sbprintf    ->  _vfiprintf_r
```

Here we see that fiprintf is called by __assert_func. call_explorer.py could
then be ran again, targetting __assert_func to walk up the calling stack until
the culprit is found.

I do not know if it is useful to include it in its own `setup.py`, given that it is all contained in a single file with no dependencies. 😐

### DIFF
--- a/tools/call_explorer.py
+++ b/tools/call_explorer.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3
+"""
+Finds all the callers for a given function, given a disassembly (list) file.
+This can be useful, i.e. when tracking down why a binary is getting very large.
+
+The callee name is given as a regular expression, which is useful when the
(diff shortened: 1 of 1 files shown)


SyrianSpock:
LGTM, we can merge it here.

PR #24, opened 2017-01-17
Calibrate / Homologate Debra

SyrianSpock:
This PR introduces the following changes:
- Robot config is handled through a yaml file, one separate for each robot
- Calibrated wheelbase parameters for Debra
- Support of front or back side calibration of the wheelbase
- Support a configurable beacon setup

Fixes introduced introduced:
- Remove config when running `make clean` 43ba18a
- Input filtering activated on encoder interface as we have noticed spurious episodes of malfunction on the right encoder interface (not the sensor) 	0291a4a
- Ignore out of date beacon signal, otherwise the robot would think the opponent is colliding with the robot if there is no beacon signal (so it wouldn't move if there was no detected beacon) a7a34dd


master-firmware/Makefile
```diff
@@ -52,6 +52,13 @@ ifeq ($(USE_BOOTLOADER),)
   USE_BOOTLOADER = no
 endif
 
+# Config file depending on robot
+ifeq ($(DEBRA),)
```
antoinealb: Can you document this option in the readme please ?

master-firmware/src/base/base_controller.c
```diff
@@ -39,14 +38,21 @@ void robot_init(void)
...
+    rs_set_right_ext_encoder(
+        &robot.rs, cvra_encoder_get_right_ext, NULL,
+        config_get_scalar("master/odometry/right_wheel_correction_factor"));
 
     /* Position manager */
     position_init(&robot.pos);
     position_set_related_robot_system(&robot.pos, &robot.rs); // Link pos manager to robot system
 
-    position_set_physical_params(&robot.pos, ROBOT_EXTERNAL_TRACK_LENGTH_MM, EXTERNAL_ENCODER_TICKS_PER_MM);
+    position_set_physical_params(
```
antoinealb: indentation seems messed up here, is this what `uncrustify`  outputs ?

master-firmware/src/base/base_controller.h
```diff
@@ -59,6 +59,7 @@ struct _robot {
 
     enum board_mode_t mode;                 // The current board mode
 
+    int calibration_direction;              // 1 if calibration side in front, -1 if in back
```
antoinealb: Maybe replace it with an enum ?

master-firmware/src/can/uavcan_node.cpp
```diff
@@ -241,8 +240,8 @@ void main(void *arg)
                           sizeof(proximity_beacon_topic_value));
 
     messagebus_advertise_topic(&bus, &proximity_beacon_topic, "/proximity_beacon");
-    const float reflector_radius = 0.040f;
-    const float angular_offset = M_PI / 2.;
+    static float reflector_radius = config_get_scalar("master/beacon/reflector_radius");
```
antoinealb: Why so `static` ?

master-firmware/src/base/base_controller.c
SyrianSpock: There is no uncrustify config

antoinealb: There is one in cvra/coding style. But I agree that it may not be the best place.

master-firmware/Makefile
SyrianSpock: Done in ad54d91c25cf824ab14b449ac67476c583bd21bb

PR #26, opened 2017-02-15
Basic support of Debra's arms

SyrianSpock:
This PR implements basic support of Debra's arms:
- Arm kinematics using code form 2014.
- Arms can be moved, given position setpoints.
- Index positions of arms motors can automatically be determined. There is a homing sequence that implements this and runs it before the wheelbase auto positioning. Homing is done in parallel for all arm motors.
- Run a different config/strategy depending on the robot (Debra/Sandoi).

Not included:
- Z axis support, not ready mechanically (need a homing switch)
- Hand support will come in a separate PR


master-firmware/src/arms/arms_controller.c
```diff
@@ -0,0 +1,86 @@
...
+
+
+static THD_FUNCTION(arms_ctrl_thd, arg)
+{
+    (void) arg;
+    chRegSetThreadName(__FUNCTION__);
+
+    while (1) {
+        /* Wait 10 milliseconds (100 Hz) */
+        chThdSleepMilliseconds(1000 / ARMS_FREQUENCY);
```
antoinealb: Empty thread, is that normal ?

master-firmware/tests/scara.cpp
```diff
@@ -0,0 +1,330 @@
...
+{
+    scara_trajectory_append_point(&traj, 10, 10, 10, COORDINATE_ARM, 1.);
+    scara_trajectory_append_point(&traj, 10, 10, 10, COORDINATE_ARM, 10.);
+
+    scara_do_trajectory(&arm, &traj);
+    CHECK_EQUAL(traj.frame_count, arm.trajectory.frame_count);
+    CHECK(0 == memcmp(traj.frames, arm.trajectory.frames, sizeof(scara_waypoint_t) * traj.frame_count));
+}
+
+// IGNORE_TEST(ArmTestGroup, ExecuteTrajectoryIsAtomic)
```
antoinealb: So no atomicity for now?

master-firmware/src/arms/arms_controller.c
SyrianSpock: Good catch, I'll remove that.

master-firmware/tests/scara.cpp
SyrianSpock: Not even using the trajectory execution yet.

PR #27, opened 2017-03-16
CAN IO code for hands

antoinealb:


master-firmware/src/main.c
```diff
@@ -268,6 +268,15 @@ void init_base_motors(void)
     motor_manager_create_driver(&motor_manager, "right-wheel");
 }
 
+
+void init_hands(void)
+{
+    bus_enumerator_add_node(&bus_enumerator, "left-fingers", NULL);
+    bus_enumerator_add_node(&bus_enumerator, "left-sensors", NULL);
+    bus_enumerator_add_node(&bus_enumerator, "right-fingers", NULL);
```
antoinealb: @nuft why is there both left-sensors and left-fingers ?

nuft: initially, we planned two io-boards per hand: one for controlling the fingers and one for the sensors

nuft: good catch, we can update this to 'left-hand', 'right-hand'.

antoinealb: Oh right, I thought it was in a recent commit.

PR #28, opened 2017-03-22
Add a tool to inspect ChibiOS using GDB

antoinealb:
So this is some code that works for me and I thought it could be useful to share it with other members of the club.

This piece of code parses ChibiOS' internal data structures to display information about the state of every thread. It can be used from interrupts and panic situations as well, although the data structures might be corrupted.

### DIFF
--- a/tools/Readme.md
+++ b/tools/Readme.md
@@ -1,5 +1,13 @@
 # Tools help
 
+# ChibiOS GDB integration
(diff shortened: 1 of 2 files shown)


antoinealb:
So apparently OpenOCD has ChibiOS integration for a while now, so let's remove this.

PR #31, opened 2017-03-27
UAVCAN cleanup

antoinealb:
This PR contains a huge cleanup and modularization of the UAVCAN code for the motor board firmware. It has not yet been tested on hw, I will do that on Wednesday.

Highlights:

* Motor board parameters should now be exposed in the UAVCAN GUI :robot:
* No more huge functions, clear separations of concerns.
* No UAVCAN data structure lives on the stack anymore, less threat of a stack overflow 🍓  

I will test all of this on Wednesday, and hopefully merge it. My next plans for UAVCAN are to add the ability to stream log output on the CAN bus.


nuft:
The size of `uavcan_node.cpp` got a bit out of hand and this cleanup was really needed.

> Motor board parameters should now be exposed in the UAVCAN GUI

Cool, I didn't know about the uavcan::ParamServer.

The PR looks good to me, good job.

motor-control-firmware/src/uavcan/uavcan_node.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+static int uavcan_node_start(Node &node)
+{
+    return node.start();
+}
+
+/** Start all UAVCAN services. */
+static void uavcan_services_start(Node &node)
+{
+    struct {
```
antoinealb: Good catch, fixed it.

SyrianSpock:
LGTM & works on the robot

PR #32, opened 2017-03-28
WIP: Send logs over UAVCAN

antoinealb:


antoinealb:
@SyrianSpock If you have some time to test it tonight, I would be interested in the results. I added this feature to both the CANIO and motor board.

PR #33, opened 2017-03-29
Deprecate StringID in favor of UAVCAN's standard

antoinealb:
UAVCAN already integrates a human readable name discovery mechanism in
the form of the NodeInfo. The idea is that when a node status with an
unknown ID is discovered, the master will send a GetNodeInfo request to
this node, which contains the name of the device. This is the mechanism
used by the GUI for example.

Libuavcan already contains a mechanism to deal with this protocol logic
and just needs a small adapter class to adapt to our application logic.


SyrianSpock:
LGTM

PR #35, opened 2017-04-03
Expose most motor board parameters

antoinealb:
Now the only things left are the stream frequencies, on which I am working as well. But I think we can merge it now.


SyrianSpock:
LGTM

PR #36, opened 2017-04-04
Port to use Nucleo144 (breaks Olimex E407)

nuft:
Start playing around with the Nucleo144 and shield.

Things that need to be tested:
- CAN
- SDCard
- Ethernet
- ServoPWM
- Encoders

Things that need to be fixed:
- the debug UART changed: UART7 is not supported (maybe upgrade ChibiOS?)
- choose a pin assignment for the EXT_IO connector


antoinealb:
I don't know how much I like the idea of upgrading ChibiOS. It is definitely something we need to do, but I fear hidden bugs such as the tickless mode enabled by default...

antoinealb:
It will also require updating UAVCAN, but I fear that less.

nuft:
Tested so far:

- CAN
- Ethernet
- Encoders left and right (transversal not yet)

Issues:
- SDCard works on only 1 out of 3 boards.

antoinealb:
We got an sdcard sniffing adapter at WISE, might be useful to debug your issues with the sdcard.

SyrianSpock:
LGTM
The build fails on drone, same as on my computer, because of the update to arm gcc 6.3 which should be investigated. I think we should merge this ASAP, it's been dragging on for too long.

SyrianSpock:
Do we need to support Olimex E407 ? Currently both robots are equipped with the NucleoF429.

If we want that, then we need to:
- have different board configurations for each board
- introduce a flag to build for a specific board
- have different UART (and other peripheral) settings for each board
- disable some features depending on the board

antoinealb:
>  Currently both robots are equipped with the NucleoF429.

Ok, we can merge then. We will investigate the build breakage later.

antoinealb:
test github hook

Issue #22, opened 2016-12-21
Support two opponents (only one supported for now)


SyrianSpock:
We only handle one opponent in the map and the dynamic avoidance.
It would be good to support two opponents (i.e. distinguish two beacon signals from two different reflectors and so on).

antoinealb:
What is the status on this ? We will definitely need that one for Belgium cup.

@SyrianSpock

antoinealb:
Relevant PR: #39

PR #39, opened 2017-04-11
Add support of two opponents in the map/obstacle avoidance

SyrianSpock:
This PR solves #22.

### DIFF
--- a/master-firmware/src/strategy.cpp
+++ b/master-firmware/src/strategy.cpp
@@ -42,6 +42,12 @@ static void wait_for_autoposition_signal(void)
     wait_for_starter();
 }
 
+void strategy_stop_robot(struct _robot* robot)
+{
+    trajectory_hardstop(&robot->traj);
+    chThdSleepMilliseconds(200);
+}
+
 bool strategy_goto_avoid(struct _robot* robot, int x_mm, int y_mm, int a_deg)
 {
     /* Create obstacle at opponent position */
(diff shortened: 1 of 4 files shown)


nuft:
There is an issue if the proximity beacon update is faster than the calls to `strategy_gogo_avoid()`. The position read from messagebus might always be the same opponent while missing the second one. A solution coud be to track the opponents somewhere else and publish a list of positions.

Anyway, I think it's good enough for now and the changes LGTM. I propose to improve opponent tracking later.

master-firmware/src/strategy.cpp
```diff
@@ -42,6 +42,12 @@ static void wait_for_autoposition_signal(void)
     wait_for_starter();
 }
 
+void strategy_stop_robot(struct _robot* robot)
+{
+    trajectory_hardstop(&robot->traj);
```
antoinealb: Is this enough to stop the robot? I think we discussed that it was not working once.

master-firmware/src/strategy.cpp
```diff
@@ -69,6 +75,11 @@ bool strategy_goto_avoid(struct _robot* robot, int x_mm, int y_mm, int a_deg)
     point_t *points;
     int num_points = oa_get_path(&points);
     DEBUG("Path to (%d, %d) computed with %d points", x_mm, y_mm, num_points);
+    if (num_points <= 0) {
+        WARNING("No path found!");
```
antoinealb: Good enough for now, but I think we might want to start a recovery strategy here instead.

PR #48, opened 2017-05-02
Remove deprecated trajectory control mode

antoinealb:
The trajectory control mode was intended for use with a PC streaming
setpoints to the master board, which would then interpolate them for the
motor boards. It is not needed anymore in the current system
architecture.


antoinealb:
Thans for spotting the unused defines! Do you think there is a nede for hardware testing of this ?

nuft:
no hardware test needed because not used.

PR #49, opened 2017-05-02
Manually manage reentrancy for printf

antoinealb:
It appears that most Newlib functions are not thread safe, and this may
cause problems in some cases. To prevent this, we can use the reentrant
variants manually.

Another option is to recompile Newlib with the __DYNAMIC_REENT__ flag
defined and implement the __getreent syscalls. However it is not really
convenient to ask the whole team to build Newlib locally.


antoinealb:
So I tested this on hardware, but could not reproduce the crash without it. @nuft can you test with and without this patch and confirm that it fixes your issue?

PR #50, opened 2017-05-10
PID tune shell command

nuft:
Tool to make tuning PIDs from shell a bit less of a pain.


antoinealb:
Looks good, nice job :)

PR #51, opened 2017-05-15
5th axis support

SyrianSpock:
This PR introduces support of the new wrist on Debra with the added 5th axis (pitch).

The PR covers:
- cleanup of arm motor callbacks
- cleanup of arm motor indexing
- indexing routine for wrist assemblies (two differential motors for heading and pitch)
- update the jacobian to include the pitch axis
- control pitch (set to position 0 for now, support in trajectories is coming soon)

Right arm is momentarily disabled.

### DIFF
--- a/master-firmware/src/arms/cvra_arm_motors.c
+++ b/master-firmware/src/arms/cvra_arm_motors.c
@@ -1,3 +1,4 @@
+#include <math.h>
 #include <error/error.h>
 #include "main.h"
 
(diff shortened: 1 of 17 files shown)


master-firmware/src/arms/cvra_arm_motors.c
```diff
@@ -1,166 +1,96 @@
...
+    /* Bring pitch between -PI and PI */
+    while (*pitch >= M_PI) {
+        *pitch -= 2 * M_PI;
+    }
+    while (*pitch <= - M_PI) {
+        *pitch += 2 * M_PI;
+    }
 
-    return dev->direction * (motor_driver_get_and_clear_stream_value(motord, MOTOR_STREAM_POSITION) - dev->index);
+    NOTICE("Up %.3f Down %.3f Heading %.3f Pitch %.3f", up, down, *heading, *pitch);
```
antoinealb: Please replace by `DEBUG`, otherwise its a mess ;)

SyrianSpock: good catch, thanks

Issue #53, opened 2017-08-21
Build fails on ARM GCC 6.3.1


SyrianSpock:
http://178.32.216.117:8000/cvra/robot-software/294

SyrianSpock:
Updated ChibiOS, doesn't seem to fix it.
[This thread](http://www.chibios.com/forum/viewtopic.php?t=1839) suggests it's related to the `exit` function being called somewhere.

antoinealb:
I am a bit surprised that this is triggered by an update. Did newlib change at the same time or something ?

nuft:
in [newlib fini.c](https://github.com/bminor/newlib/blob/e0f24404b3fcfa2c332ae14c3934546c91be3f42/newlib/libc/misc/fini.c#L20)  the function `_fini()` is declared as follows:
```
extern void _fini (void);
```
It is called in `__libc_fini_array()` after the cleanup routines.

Just defining `_fini()` gets rid of the liker issue. I didn't find out what it's actually supposed to do. Though in case of master-firmware we will never call `exit()` so it shouldn't be a problem.

What's strange is that this problem appears only now. Where was `_fini()` defined before?

antoinealb:
> What's strange is that this error appears only now. Where was _fini() defined before?

`_fini` was not called before:

```
$  arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for ARM Embedded Processors) 5.4.1 20160919 (release) [ARM/embedded-5-branch revision 240496]
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
$ ../tools/call_explorer.py build/ch.list _fini
trajectory_finished  -> trajectory_angle_finished
trajectory_has_ended -> trajectory_finished
```

@SyrianSpock could you take a look at who's calling `_fini` using `call_explorer.py` please ?

antoinealb:
I think this might be due to exceptions, because `exit` is called by `abort`.

SyrianSpock:
As @nuft mentioned `_fini` is called by `__libc_fini_array` which is called after the end of `main()`. [Relevant stackoverflow thread](https://stackoverflow.com/questions/15265295/understanding-the-libc-init-array).
```
$ arm-none-eabi-gcc --version
[...]

PR #55, opened 2017-09-11
Fix build for GCC 6.3

nuft:
Addresses #53


antoinealb:
Could you add it to `chibios-syscalls` instead to avoid code dup ?

SyrianSpock:
Tested on hardware the master board, motor boards, and proximity beacon, they all seem to behave properly.

SyrianSpock:
@antoinealb chibios-syscalls is only used by can-io-firmware, the other use syscalls defined by ChibiOS. This refactoring should be done in another PR to properly test the changes (see #57). For the time being I'll define _fini in chibios-syscalls for can-io-firmware.

PR #58, opened 2017-09-26
Update to ChibiOS 4.0.0

nuft:
This PR updates all firmwares to ChibiOS 4.0.0
UAVCAN also needed an update to support the new version of ChibiOS.

All firmwares need to be tested on hardware before the merge.


antoinealb:
One important difference is that ChibiOS/RT 4.0 now includes an MPU driver to check for stack overflows, therefore we should remove our own implementation.

nuft:
> One important difference is that ChibiOS/RT 4.0 now includes an MPU driver to check for stack overflows, therefore we should remove our own implementation.

Agree. Will do once merged.

SyrianSpock:
For the issue related to stack protection. If testing fails, then we need to fix it before merging this.
Otherwise, we can fix it in a following PR.

#59 should track this.

nuft:
I did some tests on the hardware:
All boards are visible over UAVCAN.
master and can-io are working fine.
motor-control works for the wheelbase (current control)
proximity-beacon does not work (voltage control)

antoinealb:
> proximity-beacon does not work (voltage control)
 
what are the symptoms ?

nuft:
> what are the symptoms ?

The beacon is visible over UAVCAN but the motor is not turning.

antoinealb:
Thats pretty surprising, I have no idea what could create this type of issues. I mean, the platform worked completely for the wheelbase :/

May be a stupid question, but the same beacon works with old code ?

nuft:
Yes, the beacon works with master.

antoinealb:
And is it panicking or something ?

nuft:
> And is it panicking or something ?

No, it's publishing data on UAVCAN. just the motor doesn't turn. Perhaps a bug or an API change of one of the drivers.

antoinealb:
Its publishing the beacon as well? My hypothesis was that the UAVCAN timer was broken and that [this line](https://github.com/cvra/robot-software/blob/master/proximity-beacon-firmware/src/proximity_beacon_publisher.cpp#L23) was not executed

SyrianSpock:
Yes it's publishing the sensor reading.

nuft:
With @froj we found and fixed the issue with the motor-control and proximity-beacon firmware.
It was a bug in the ChibiOS ADC driver.
The latest fix works on hardware.
This PR is ready to be merged.

PR #62, opened 2017-10-26
Update lwip to latest stable

antoinealb:
Hello,
This PR updates lwip to the latest stable version (2.0.3) and re-enables the code in main for testing.  I will test it on hardware tonight.

I think it would be best to keep IP enabled as much as possible to find issues with the implementation.


antoinealb:
I also changed lwip submodule URL to point to upstream (as we are running without any patches anymore yay). What do you think? Should we keep our own fork?

SyrianSpock:
No need to keep the fork if we don't apply any patch. Git takes care of the versioning.

PR #63, opened 2017-10-28
Arm refactoring #1: Split joint

SyrianSpock:
This PR extracts a `joint_t` struct removing duplication from `scara_t`.


SyrianSpock:
@antoinealb Don't you trust our test suite well enough? :smiling_imp: but yes I tested on hardware.

antoinealb:
Actually for this type of changes, its true that the test suite might be enough.

PR #65, opened 2017-10-29
Arm refactoring #2: Remove unused stuff

SyrianSpock:
This PR removes hacks/features not used in the arm's code:
- The variable length hack introduced last year to handle the change of forearm length depending on the hand's pitch
- The kinematics solution count is only used locally, no need to store it in the arm's state
- The last loop time was previously recorded but never used for lag compensation as advertised by comments.
- The unused trajectory semaphore


SyrianSpock:
image

PR #66, opened 2017-10-30
Arm refactoring #3: Introduce Lie group formalism

SyrianSpock:
This PR introduces Lie group formalism to abstract away rigid body transformations done in the code.
Defines only 2D rotations (SO2) and 2D transforms (SE2).

### DIFF
--- a/master-firmware/src/scara/lie_groups.c
+++ b/master-firmware/src/scara/lie_groups.c
@@ -20,6 +20,11 @@ point_t so2_rotate(so2_t rotation, point_t point)
     return point;
 }
 
+point_t so2_inverse_rotate(so2_t rotation, point_t point)
+{
+    return so2_rotate(so2_create(- rotation.angle), point);
+}
+
 vect_t translation_2d(float x, float y)
 {
     vect_t translation = {.x = x, .y = y};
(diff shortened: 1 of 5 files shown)


master-firmware/src/scara/lie_groups.h
```diff
@@ -0,0 +1,31 @@
...
+// Translation
+vect_t translation_2d(float x, float y);
+
+// Special Euclidian group 2 (SE2) defines a 2D rigid body transform (rotation + translation)
+typedef struct {
+    so2_t rotation;
+    vect_t translation;
+} se2_t;
+
+se2_t se2(float angle, vect_t translation);
```
antoinealb: Can we rename this into something like `se2_create` or `se2_make` ? It would make the code easier to search.

master-firmware/src/scara/lie_groups.c
```diff
@@ -0,0 +1,58 @@
+#include "lie_groups.h"
+
+so2_t so2(float angle)
```
antoinealb: Maybe we should rename this to `so2_create `to make it easier to find ?

SyrianSpock: Sure. I miss my C++ constructors :stuck_out_tongue:

master-firmware/src/scara/lie_groups.h
SyrianSpock: Done

PR #67, opened 2017-11-08
Arm refactoring #4: Some abstraction cleanup 

SyrianSpock:
This PR extracts some code into functions to make functions shorter / code more readable. 
UTs are also cleaned up.

### DIFF
--- a/master-firmware/src/scara/scara.c
+++ b/master-firmware/src/scara/scara.c
@@ -89,6 +89,13 @@ void scara_do_trajectory(scara_t *arm, scara_trajectory_t *traj)
     chMtxUnlock(&arm->lock);
 }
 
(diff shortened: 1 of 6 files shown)


master-firmware/src/scara/scara.c
```diff
@@ -8,6 +8,18 @@
 #include "scara_port.h"
 #include "scara_jacobian.h"
 
+static void scara_shutdown_joints(scara_t* arm);
+
+static void scara_lock(mutex_t* mutex)
```
antoinealb: [question] Any reason to put this in a separate function? It does not improve portability or anything?

SyrianSpock: You're right, It should probably move to `scara_port` to improve portability. I just found it was more readable.

PR #68, opened 2017-11-12
Arm refactoring #5: Hardware interface

SyrianSpock:
This PR starts splitting the scara code into blocks. The 1st block we introduce is `scara_hardware_interface` that is handling joint IO (reading positions, setting positions and velocities).


nuft:
LGTM

PR #69, opened 2017-11-13
Arm trajectory pause/continue

SyrianSpock:
This PR introduces the ability to pause a trajectory during execution and restore it.
This should ease the process of tuning and debugging arm trajectories.
Higher level pause/continue can also be implemented using this.

Implemented in pair programming session with @nuft.

### DIFF
--- a/master-firmware/src/commands.c
+++ b/master-firmware/src/commands.c
@@ -1056,6 +1056,30 @@ static void cmd_scara_traj(BaseSequentialStream *chp, int argc, char *argv[])
     }
 }
 
+static void cmd_scara_pause(BaseSequentialStream *chp, int argc, char *argv[])
+{
+    (void)argc;
+    (void)argv;
+
+    scara_t* arm = &main_arm;
+
+    scara_pause(arm);
+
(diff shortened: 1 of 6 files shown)


SyrianSpock:
@antoinealb It would certainly be more efficient, the problem is that our current implementation works with trajectories, not single waypoints, which means we need to change `scara_position_for_date`. But that seems feasible, maybe something to try.

antoinealb:
Ah yeah I see. Nevermind then, its not like we are RAM bound for now.

antoinealb:
That said, the tests are failling on Travis CI, could you take a look ?

SyrianSpock:
@antoinealb I know, there is a weird behavior that we observed and tried to fix in e9e9d9b. I need more time to investigate, might be revealing deeper problems

SyrianSpock:
Should be fixed now. The `coordinate_type` field was not initialized in `scara_trajectory_interpolate_waypoints`. Thanks ` valgrind --leak-check=full --track-origins=yes ./tests`

PR #71, opened 2017-11-20
Add compiler flag to output stack usage per function

SyrianSpock:
This uses the `-fstack-usage` option introduced in GCC 6.1. The output is one `.su` file per module in `build/obj`, and it shows the maximum stack usage for each function.

Example:
```
strategy_helpers.c:17:6:strategy_auto_position          48    static
strategy_helpers.c:67:6:strategy_align_y                40    static
strategy_helpers.c:85:10:strategy_set_arm_trajectory    56    static
```

### DIFF
--- a/master-firmware/Makefile
+++ b/master-firmware/Makefile
@@ -6,7 +6,7 @@
 # Compiler options here.
 ifeq ($(USE_OPT),)
   USE_OPT = -O2 -ggdb -fomit-frame-pointer -falign-functions=16
-  USE_OPT += -frename-registers -freorder-blocks -fconserve-stack
+  USE_OPT += -frename-registers -freorder-blocks -fconserve-stack -fstack-usage
   USE_OPT += -fstack-protector-all -L .
 endif
 


antoinealb:
Do you think this could be used for static stack usage analysis ? Is it reliable ?

SyrianSpock:
I think it's reliable, and we can use it for static stack analysis.

From [GCC doc](https://gcc.gnu.org/onlinedocs/gcc/Developer-Options.html#Developer-Options)
```
-fstack-usage

Makes the compiler output stack usage information for the program, on a per-function basis. The filename for the dump is made by appending .su to the auxname. auxname is generated from the name of the output file, if explicitly specified and it is not an executable, otherwise it is the basename of the source file. An entry is made up of three fields:

The name of the function.
A number of bytes.
One or more qualifiers: static, dynamic, bounded.
The qualifier static means that the function manipulates the stack statically: a fixed number of bytes are allocated for the frame on function entry and released on function exit; no stack adjustments are otherwise made in the function. The second field is this fixed number of bytes.

The qualifier dynamic means that the function manipulates the stack dynamically: in addition to the static allocation described above, stack adjustments are made in the body of the function, for example to push/pop arguments around function calls. If the qualifier bounded is also present, the amount of these adjustments is bounded at compile time and the second field is an upper bound of the total amount of stack used by the function. If it is not present, the amount of these adjustments is not bounded at compile time and the second field only represents the bounded part.
```

PR #72, opened 2017-11-21
First version of a higher level arm API

SyrianSpock:
This PR introduces a higher level wrapper to arm trajectories `ArmTrajectory`.
It provides a fluent interface, a typical usage looks like

```c++
const auto trajectory = ArmTrajectory(&arm)
                            .startAt({1, 2, 3, COORDINATE_ARM})
                            .goThrough({4, 5, 6, COORDINATE_ROBOT})
                            .goThrough({7, 8, 9, COORDINATE_TABLE});
trajectory.execute();
wait_s(trajectory.duration());
```

And trajectories can be chained as follow
```c++
const auto trajectory = ArmTrajectory(&arm)
                            .startAt({1, 2, 3, COORDINATE_ARM})
                            .goThrough({4, 5, 6, COORDINATE_ROBOT})
                            .goThrough({7, 8, 9, COORDINATE_TABLE});
const auto finalFrame = trajectory.execute();

// do other stuff

const auto secondTrajectory = ArmTrajectory(&arm)
                                .startAt(finalFrame)
                                .goThrough({4, 5, 6, COORDINATE_ROBOT})
                                .goThrough({7, 8, 9, COORDINATE_TABLE});
```


master-firmware/src/arms/arm.h
```diff
@@ -0,0 +1,34 @@
+#ifndef ARM_API_H
```
antoinealb: Maybe we should name C++ headers files with ".hpp".

SyrianSpock: Not sure what we should have as standard

SyrianSpock:
The idea is to try out this interface and see what are the pain points. This version is just C wrapped in a class to have a fluent interface (syntactic sugar).

nuft:
> More C++ might be dangerous though, is this a tradeoff that we want ?

I don't see this as problematic, especially if we keep things simple.
In case of the strategy code it makes sense to use C++. It should be clean and easy to write.

PR #73, opened 2017-11-30
Servo driver for PCA9685 PWM controller

nuft:


master-firmware/src/servo.c
```diff
@@ -0,0 +1,117 @@
...
+    mode = pca9685_read_reg(PCA9685_MODE1);
+    pca9685_write_reg(PCA9685_MODE1, mode); // clear RESTART bit
+
+    // push-pull output, update on I2C STOP,
+    pca9685_write_reg(PCA9685_MODE2, PCA9685_MODE2_OUTDRV);
+
+    // all servo output off
+    unsigned int i;
+    for (i = 0; i < SERVO_COUNT; i++) {
+        uint8_t buf[5] = {
```
antoinealb: Is this the same as `servo_set(i, 0)`? if yes, I suggest replacing it to be more readable

nuft: It's not the same. It forces the output to 0.
Perhaps it's makes more sense to put all servos to neutral position? `servo_set(i, 0.5)`

master-firmware/src/servo.c
```diff
@@ -0,0 +1,117 @@
...
+
+    uint8_t buf[3] = {
+        PCA9685_LED0_OFF_L + 4 * servo_nb,
+        (uint8_t) count,
+        (uint8_t) (count >> 8)
+    };
+
+    i2cAcquireBus(&I2CD1);
+    i2cMasterTransmit(&I2CD1, PCA9685_I2C_ADDR, buf, sizeof(buf), NULL, 0);
+    i2cReleaseBus(&I2CD1);
```
SyrianSpock: I would use the `pca9685_read_reg`/`pca9685_write_reg` functions instead of calling the i2c driver, to separate the abstraction levels.

master-firmware/src/servo.c
```diff
@@ -0,0 +1,117 @@
...
+        pos = 1.0f;
+    } else if (pos < 0.0f) {
+        pos = 0.0f;
+    }
+
+    // 1 - 2ms pulse length
+    uint16_t count = (uint16_t)((pos + 1.0f) * (PWM_CLOCK / 1000));
+
+    uint8_t buf[3] = {
+        PCA9685_LED0_OFF_L + 4 * servo_nb,
```
SyrianSpock: I see `address + 4 * number` repeated, maybe wrap it in a function call, something like `servo_get_register_for_element(uint32_t base_register, uint32_t element)`

master-firmware/src/servo.c
SyrianSpock: I think setting PWM to 0 is fine

SyrianSpock:
Covers #79

nuft:
I implemented most changes proposed by @SyrianSpock 
Needs testing on HW

PR #80, opened 2017-12-07
Support lever module

SyrianSpock:
This PR wraps the lever module (2 pumps + 1 servo) in one module that allows to
- turn pumps on/off
- deploy/retract the lever

This should be enough to start using it.

Tested on hardware.

### DIFF
--- a/master-firmware/src/lever/lever.c
+++ b/master-firmware/src/lever/lever.c
@@ -2,18 +2,34 @@
 
 #include <string.h>
 
+static void lever_lock(mutex_t* mutex)
+{
+    chMtxLock(mutex);
+}
+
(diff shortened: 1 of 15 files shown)


master-firmware/src/lever/lever.c
```diff
@@ -0,0 +1,52 @@
+#include "lever.h"
+
+#include <string.h>
+
+static const float LEVER_SERVO_RETRACTED_POS = 0.9f * 0.001f;
```
antoinealb: Maybe put this in a parameter ?

SyrianSpock: good idea, i'll do that in this PR

PR #81, opened 2017-12-17
Motor board: store config in flash

antoinealb:
This commit implements the first part of the code required for issue #3 . Now we can store the config into flash using the UAVCAN GUI and it will be automatically loaded on boot, making the board ready to use.

I tested it on hardware, but could not test it with a motor.


motor-control-firmware/linker/STM32F303xC.ld
```diff
@@ -20,7 +20,8 @@
  */
 MEMORY
 {
-    flash0  : org = 0x08000000, len = 256k
+    flash0  : org = 0x08000000, len = 252k
```
SyrianSpock: Shouldn't this be 254k?

motor-control-firmware/src/uavcan/uavcan_node.cpp
```diff
@@ -111,6 +116,11 @@ static THD_FUNCTION(uavcan_node, arg)
             uavcan_failure("UAVCAN spin");
         }
 
+        if (chBSemWaitTimeout(&node_init_complete, TIME_IMMEDIATE) == MSG_OK) {
```
SyrianSpock: Cool

SyrianSpock:
Test on hardware:
- Motor boards seem to work
- Proximity beacon seems to panic (I don't even think the proximity beacon cares about the config, it runs no control)
- Master board panics (UAVCAN service timeout when trying to load config to boards that already have one)

It also makes me wonder if we shouldn't move the control code out of motor board, so that we can use it to write fully separe apps between proximity beacon and the motor board, but that's another discussion.

antoinealb:
> Motor boards seem to work

> Proximity beacon seems to panic

Weird, not sure why, will debug.

> Master board panics (UAVCAN service timeout when trying to load config to boards that already have one)

As expected I suppose ? Although it might be a problem for merging right now.

antoinealb:
Pushed a fix for the proximity beacon, needs testing on hardware.

antoinealb:
Tested on hardware for proximity beacon and appears to work.

SyrianSpock:
Tested on robot, beacon works with the fix. OK to merge.

Issue #74, opened 2017-12-01
CAN-IO support for ramp PWM output to protect servo


nuft:

antoinealb:
Basic idea of how it would work:

* The UAVCAN message (`20131.ServoPWM.uavcan`) would be extended with a speed & acceleration variable.
* The quadramp module will be used for ramp generation.
* As we don't have enough RAM for additional stack, it will probably run from the UAVCAN thread, using an UAVCAN timer.

Does this work for you guys ?

SyrianSpock:
Sounds good, it can still be wrapped in a separate module (that depends on quadramp) and exposes a simple interface that's used inthe UAVCAN thread. Could also be used on the master board.

PR #82, opened 2018-01-02
CAN-IO: Support ramps on the servo outputs

SyrianSpock:
This PR addresses issue #74, the support for ramps using the quadramp module is added to the servo outputs on the CAN-IO board. 1st and 2nd order coefficients are set via the UAVCAN ServoPWM message, a coefficient set to zero is ignored and the previous value is kept.


can-io-firmware/set_servo.py
```diff
@@ -0,0 +1,58 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "port",
+        help="SocketCAN interface (e.g. can0) or SLCAN serial port (e.g. /dev/ttyACM0)"
+    )
+
+    parser.add_argument("id", help="ID of the board to target", type=int)
+
+    parser.add_argument("servo", help="Servo output to be controlled", type=int)
+    parser.add_argument("pos", help="Desired duty cycle on the servo output", type=float)
```
antoinealb: Maybe provide default values in the script ?

can-io-firmware/src/pwm.h
```diff
@@ -0,0 +1,27 @@
...
+{
+    PWM_CHANNEL_0, // Timer 16 channel 0
+    PWM_CHANNEL_1, // Timer 17 channel 0
+    PWM_CHANNEL_2, // Timer 1 channel 1
+    PWM_CHANNEL_3, // Timer 1 channel 2
+};
+
+void pwm_init(uint32_t frequency, uint32_t period);
+
+void pwm_set_pulsewidth(enum pwm_channel channel, uint32_t pulsewidth);
```
antoinealb: Maybe add a comment / variable name to give the units of period and pulsewidth ? Is it ms or clock ticks ?

can-io-firmware/src/servo.c
```diff
@@ -0,0 +1,93 @@
...
+        quadramp_set_1st_order_vars(filter, vel * SERVO_PWM_TIMER_FREQ, vel * SERVO_PWM_TIMER_FREQ);
+    }
+    if (acc > 1e-7) {
+        quadramp_set_2nd_order_vars(filter, acc * SERVO_PWM_TIMER_FREQ, acc * SERVO_PWM_TIMER_FREQ);
+    }
+}
+
+void servo_set(const float pos[4], const float vel[4], const float acc[4])
+{
+    servo[0].setpoint = pos[0];
```
antoinealb: Please add proper locking since `servo` is shared between the UAVCAN thread and the servo_thd.

lib/quadramp/tests/quadramp.cpp
```diff
@@ -0,0 +1,30 @@
+#include "CppUTest/TestHarness.h"
```
antoinealb: Thanks for writing a smoke test for this module 👍

can-io-firmware/src/servo.c
```diff
@@ -0,0 +1,93 @@
...
+    } else if (pos < 0) {
+        pos = 0;
+    }
+    return (uint32_t)(pos * SERVO_PWM_TIMER_FREQ);
+}
+
+/* Not set if coefficient is too small (close to zero) */
+static void set_quadramp_coefficients(struct quadramp_filter* filter, float vel, float acc)
+{
+    if (vel > 1e-7) {
```
antoinealb: Could be as well > 0 and just say that negative values are "Do not change".

uavcan_data_types/cvra/io/20131.ServoPWM.uavcan
```diff
@@ -6,3 +6,5 @@
 uint7 node_id
 
 float16[4] servo_pos
+float16[4] servo_vel
```
antoinealb: Maybe document the fact that negative values are interpreted as "Do not change".

can-io-firmware/src/servo.c
SyrianSpock: Good catch!

can-io-firmware/src/servo.c
SyrianSpock: Yep

can-io-firmware/src/pwm.h
SyrianSpock: Clock ticks, adding a comment

SyrianSpock:
Thanks for the review. I only tested it by writing the filtered servo setpoint to the blinking LED, still need to test on real servos.

can-io-firmware/set_servo.py
SyrianSpock: Done

SyrianSpock:
Tested on servos, and works well.

PR #85, opened 2018-01-12
Add a tool to configure all motor boards at once

antoinealb:
This PR adds tool to read the `config.yml` file and write the config into all motor boards at once. Allows us to keep the robot's config in Git, even if the boards can now be configured by hand. Its like "saltstack for motor boards".


SyrianSpock:
Fixes #84

antoinealb:
Nope, not tested on HW.

PR #86, opened 2018-01-19
CVRA Studio - MVP 1: Node discovery

SyrianSpock:
Basic tool performing node discovery over CAN.

Example usage:
```
➤ python node_discovery.py slcan0 --dsdl ../../uavcan_data_types/cvra/
ID    Name                 Status               Health               Uptime    
   45 demo                 OPERATIONAL          OK                   0:00:02   
  120 test                 INITIALIZATION       OK                   0:00:03   
```

The features are very basic, but it shows the idea of what we want to build and how: Model View Controller architecture and common UAVCAN interface handling spin/lock.

### DIFF
--- a/tools/studio/node_discovery.py
+++ b/tools/studio/node_discovery.py
@@ -1,0 +1,19 @@
+import threading
+import time
+import uavcan
+
+def uavcan_thread(node):
+    while True:
+        try:
+            node.spin(0.1)
+            time.sleep(0.01)
+        except uavcan.UAVCANException as ex:
+            print('Node error:', ex)
+            return
+
+def main():
+    node = uavcan.make_node("slcan0", node_id=127)
+    threading.Thread(target=uavcan_thread, args=(node,)).start()
+
+if __name__ == '__main__':
+    main()
--- a/tools/studio/node_discovery.py
+++ b/tools/studio/node_discovery.py
@@ -27,16 +27,61 @@ class UavcanNode:
         for handler in self.handlers:
             handler.remove()
 
+class NodeStatusModel:
+    def __init__(self, node):
+        self.known_nodes = {}
+        node.add_handler(uavcan.protocol.NodeStatus, self._node_status_callback)
+
+    def _node_status_callback(self, event):
+        self.known_nodes[event.transfer.source_node_id] = event.message
+
+class NodeStatusViewer:
+    def __init__(self):
+        self.status_messages = {
+            uavcan.protocol.NodeStatus().MODE_OPERATIONAL: 'OPERATIONAL',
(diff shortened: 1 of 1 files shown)


SyrianSpock:
Not sure if I should also add this script that spins a dummy node on UAVCAN, useful to debugging

```python
import argparse
import uavcan

def parse_args():
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("interface", help="Serial port or SocketCAN interface")
    parser.add_argument("id", help="UAVCAN node ID", type=int)
    parser.add_argument("name", help="UAVCAN node name", type=str)

    return parser.parse_args()

def main():
    args = parse_args()

    node_info = uavcan.protocol.GetNodeInfo.Response(name=args.name)
    node = uavcan.make_node(args.interface, node_id=args.id, node_info=node_info)

    while True:
        try:
            node.spin(1)
        except uavcan.UAVCANException as ex:
            print('Node error:', ex)

if __name__ == '__main__':
    main()
```

Usage: `python node.py slcan0 120 test`

tools/studio/node_discovery.py
```diff
@@ -0,0 +1,134 @@
...
+class UavcanNode:
+    def __init__(self, interface, node_id):
+        self.handlers = []
+        self.node_lock = threading.RLock()
+        self.node = uavcan.make_node(interface, node_id=node_id)
+
+    def subscribe(self, topic, callback):
+        self.handlers.append(self.node.add_handler(topic, callback))
+
+    def service_call(self, request, node_id, callback):
```
antoinealb: Maybe name it `request`, since it wraps UAVCAN's request method ?

tools/studio/node_discovery.py
```diff
@@ -0,0 +1,134 @@
...
+import time
+import uavcan
+
+class UavcanNode:
+    def __init__(self, interface, node_id):
+        self.handlers = []
+        self.node_lock = threading.RLock()
+        self.node = uavcan.make_node(interface, node_id=node_id)
+
+    def subscribe(self, topic, callback):
```
antoinealb: Same as below, we could name it add_handler.

tools/studio/node_discovery.py
```diff
@@ -0,0 +1,134 @@
...
+        print(formatted_line.format("ID", "Name", "Status", "Health", "Uptime"))
+        for node in nodes:
+            name = nodes[node]['name']
+            status = nodes[node]['status']
+            print(formatted_line.format(node, name, self._display_status(status.mode),
+                                        self._display_health(status.health),
+                                        self._display_uptime(status.uptime_sec)))
+        self._previous_print_len = 1 + len(nodes)
+
+    def _delete_previous_print(self):
```
antoinealb: How does it delete code ? Maybe drop a comment.

PR #94, opened 2018-02-25
[WIP] Stream master PID feedback over UAVCAN

SyrianSpock:


master-firmware/src/can/uavcan_node.cpp
```diff
@@ -127,6 +132,35 @@ static void main(void *arg)
...
+        ERROR("DistancePID publisher");
+    }
+    angle_pid_pub.construct<Node&>(node);
+    res = angle_pid_pub->init();
+    if (res < 0) {
+        ERROR("AnglePID publisher");
+    }
+
+    uavcan::Timer pid_feedback_timer(node);
+    pid_feedback_timer.setCallback([&](const uavcan::TimerEvent& event)
```
nuft: Perhaps don't stream the PID when not used.
Though, the functionality to enable/disable can be added later.

pavel-kirienko: I'm just a passerby here, so feel free to disregard this if you feel it doesn't concern your project. I just thought that maybe you don't know that *C++ closures may allocate heap memory at every call*. The implementation supplied with your stdlib does that certainly. Enjoy!

SyrianSpock: > C++ closures may allocate heap memory at every call

Oh, I didn't know this. Thanks for pointing it out. Some cleanup is in order.

PR #101, opened 2018-03-18
Sensor board

nuft:
This PR adds the new sensor board firmware and integrates it into master firmware where the distance readings are published on messagebus. Everything was tested on hardware.

The sensor board code needs some cleanup and it has no bootloader support yet. Therefore the UAVCAN ID is hardcoded for the moment. Nevertheless, I suggest to merge now to move forward with strategy integration.


sensor-board-firmware/package.yml
```diff
@@ -0,0 +1,25 @@
+dependency-dir: ../lib
+
+depends:
+  - chibios-syscalls
+  - cmp
+  - cmp_mem_access
+  - error
+  - crc
+  - version
+  - quadramp
```
antoinealb: Why do we need the quadramp module ?

nuft: no, good catch. this is an artefact form can-io

SyrianSpock: Copied from CAN IO firmware, can be removed

master-firmware/src/can/sensor_handler.cpp
```diff
@@ -0,0 +1,37 @@
...
+static messagebus_topic_t hand_distance_topic;
+static MUTEX_DECL(hand_distance_topic_lock);
+static CONDVAR_DECL(hand_distance_topic_condvar);
+static float hand_distance_topic_value;
+
+static void sensor_distance_cb(const uavcan::ReceivedDataStructure<cvra::sensor::DistanceVL6180X>& msg)
+{
+    if (msg.status == cvra::sensor::DistanceVL6180X::STATUS_OK) {
+        float dist = (float)msg.distance_mm / 1000.0f;
+        messagebus_topic_publish(&hand_distance_topic, &dist, sizeof(dist));
```
SyrianSpock: Cool :sunglasses:

sensor-board-firmware/src/vl6180x/vl6180x.c
```diff
@@ -0,0 +1,133 @@
...
+    // vl6180x_write_register(dev, 0x0103, 0x01); // undocumented
+    vl6180x_write_register(dev, 0x0030, 0x00);
+    // vl6180x_write_register(dev, 0x001b, 0x0a); // SYSRANGE__INTERMEASUREMENT_PERIOD
+    // vl6180x_write_register(dev, 0x003e, 0x0a); // SYSALS__INTERMEASUREMENT_PERIOD
+    // vl6180x_write_register(dev, 0x0131, 0x04); // undocumented
+    // // vl6180x_write_register(dev, 0x0011, 0x10);
+    // // vl6180x_write_register(dev, 0x0014, 0x24);
+    // // vl6180x_write_register(dev, 0x0031, 0xff);
+    // vl6180x_write_register(dev, 0x00d2, 0x01); // undocumented
+    // vl6180x_write_register(dev, 0x00f2, 0x01); // undocumented
```
SyrianSpock: I suggest removing unused code, and do it in a separate commit if you want to keep it in history.

nuft: Ok, will do. I think this is coming from different reference implementations.

PR #110, opened 2018-03-30
Map server

SyrianSpock:
This PR refactors the way we handle the map and set obstacles on it:
- The map is now handled by a dedicated thread, the map server
- The beacon signal is watched to update the obstacle we draw around an opponent
- The strategy state is watched to update the obstacle we draw around blocks of cubes


master-firmware/src/base/map_server.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "strategy/state.h"
+
+#define MAP_SERVER_STACKSIZE 1024
+
+static THD_FUNCTION(map_server_thd, arg)
+{
+    enum strat_color_t color = *(enum strat_color_t*)arg;
+    chRegSetThreadName(__FUNCTION__);
+
+    const int robot_size = config_get_integer("master/robot_size_x_mm");
```
antoinealb: Maybe move this in the while() so that it can be changed using the parameter API.

master-firmware/src/base/map_server.cpp
```diff
@@ -0,0 +1,70 @@
...
+        messagebus_topic_read(proximity_beacon_topic, &beacon_signal, sizeof(beacon_signal));
+        if (timestamp_duration_s(beacon_signal.timestamp, timestamp_get()) < TRAJ_MAX_TIME_DELAY_OPPONENT_DETECTION) {
+            float x_opp, y_opp;
+            beacon_cartesian_convert(&robot.pos, 1000 * beacon_signal.distance, beacon_signal.heading, &x_opp, &y_opp);
+            map_update_opponent_obstacle(&map, x_opp, y_opp, opponent_size * 1.25, robot_size);
+        } else {
+            map_update_opponent_obstacle(&map, 0, 0, 0, 0); // reset opponent position
+        }
+
+        messagebus_topic_wait(strategy_state_topic, &state, sizeof(state));
```
antoinealb: If, for some reason the strategy does not publish, it will not update the beacon. Maybe use a topic_read here, especially since you chThdSleep() later.

master-firmware/src/base/map_server.cpp
SyrianSpock: done

master-firmware/src/base/map_server.cpp
SyrianSpock: good point

PR #112, opened 2018-04-01
Fix goap memory use issue

antoinealb:
Basically, once we get out of free nodes, we discard the least likely
states (i.e. those with the lowest priority). This does not guarantee
convergence, but makes it more likely in most cases.


antoinealb:
@SyrianSpock I opened this against your example branch but feel free to cherry pick it somewhere else if it seems more appropriate.

PR #114, opened 2018-04-02
Build more towers

SyrianSpock:
This PR generalizes the tower building approach to enable the construction of a second tower.
The initial cube pickup is also made smarter, by allowing 4 possible positions and trying them all from closest to farthest.

### DIFF
--- a/master-firmware/src/robot_helpers/math_helpers.c
+++ b/master-firmware/src/robot_helpers/math_helpers.c
@@ -33,3 +33,16 @@ int math_clamp_value(int value, int min, int max)
         return value;
     }
 }
+
+int argmin(float* values, int num_values)
+{
+    int min_index = 0;
+    float min_value = INFINITY;
+    for (int i = 0; i < num_values; i++) {
+        if (values[i] < min_value) {
+            min_index = i;
+            min_value = values[i];
+        }
+    }
+    return min_index;
+}
--- a/master-firmware/src/strategy.cpp
+++ b/master-firmware/src/strategy.cpp
@@ -526,11 +526,15 @@ struct BuildTowerLevel : actions::BuildTowerLevel {
             return false;
         }
 
-        if (!strat_deposit_cube(MIRROR_X(m_color, x_mm - 30), y_mm - 210, level)) {
+        const int tower_x_mm = x_mm - 30;
+        const int tower_y_mm = y_mm - 210;
+        if (!strat_deposit_cube(MIRROR_X(m_color, tower_x_mm), tower_y_mm, level)) {
             WARNING("Tower building did not go as expected");
             return false;
         }
 
+        state.tower_pos[0] = tower_x_mm;
+        state.tower_pos[1] = tower_y_mm;
         state.tower_level += 1;
         return true;
     }
--- a/master-firmware/src/strategy.cpp
+++ b/master-firmware/src/strategy.cpp
@@ -489,10 +489,10 @@ struct DepositCubes : actions::DepositCubes {
             state.lever_full_right = false;
         }
         for (int i = 0; i < 5; i++) {
-            state.cubes_ready_for_construction[i] = true;
+            state.construction_zone.cubes_ready[i] = true;
             point_t cube_pos = strategy_block_pos(blocks_pose, (enum block_color)i);
-            state.cubes_pos[i][0] = cube_pos.x;
-            state.cubes_pos[i][1] = cube_pos.y;
+            state.construction_zone.cubes_pos[i][0] = cube_pos.x;
+            state.construction_zone.cubes_pos[i][1] = cube_pos.y;
         }
 
         return true;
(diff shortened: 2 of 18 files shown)


master-firmware/src/robot_helpers/math_helpers.c
```diff
@@ -33,3 +33,16 @@ int math_clamp_value(int value, int min, int max)
         return value;
     }
 }
+
+int argmin(float* values, float num_values)
```
antoinealb: num_values should be an int.

master-firmware/src/robot_helpers/math_helpers.c
```diff
@@ -33,3 +33,16 @@ int math_clamp_value(int value, int min, int max)
...
+{
+    int min_index = 0;
+    float min_value = 1e37;
```
antoinealb: [nitpick] math.h has a macro `INFINITY`, which represents the largest value in a float.

master-firmware/src/strategy.cpp
```diff
@@ -348,28 +348,45 @@ struct PickupBlocks : actions::PickupBlocks {
...
 
-        se2_t blocks_pose = se2_create_xya(MIRROR_X(m_color, x_mm), y_mm, 0);
+        se2_t cubes_pose = se2_create_xya(MIRROR_X(m_color, x_mm), y_mm, 0);
+        se2_t pickup_poses[4] = {
+            se2_create_xya(MIRROR_X(m_color, x_mm - 160), y_mm - 160, MIRROR_A(m_color, -45)),
+            se2_create_xya(MIRROR_X(m_color, x_mm + 160), y_mm - 160, MIRROR_A(m_color,  45)),
+            se2_create_xya(MIRROR_X(m_color, x_mm + 160), y_mm + 160, MIRROR_A(m_color, 135)),
+            se2_create_xya(MIRROR_X(m_color, x_mm - 160), y_mm + 160, MIRROR_A(m_color, 225)),
+        };
+        int num_possible_poses = sizeof(pickup_poses) / sizeof(pickup_poses[0]);
```
antoinealb: [nitpick] Since we are doing c++, we might use std::array here instead of the sizeof hack :wink:

master-firmware/src/strategy/state.h
```diff
@@ -17,14 +18,18 @@ struct RobotState {
...
-    int cubes_pos[5][2] = {{0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}};
-    int tower_level{0};
+    enum cube_color tower_sequence[5] = {CUBE_YELLOW, CUBE_BLACK, CUBE_BLUE, CUBE_GREEN, CUBE_RED};
+    struct {
+        bool cubes_ready[5] = {false, false, false, false, false}; // YELLOW, GREEN, BLUE, RED, BLACK
+        uint16_t cubes_pos[5][2] = {{0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}};
+        uint8_t tower_level{0};
+        uint16_t tower_pos[2] = {0, 0};
+    } construction_zone[2];
+    uint16_t construction_zone_pos[2][2] = {{500, 300}, {900, 300}};
```
antoinealb: Why do we need this in the state ? Can it change during the game ?

master-firmware/tests/test_map.cpp
```diff
@@ -213,3 +213,11 @@ TEST(AMap, canSetCubesObstacleAtomically)
 
     map_set_cubes_obstacle(&map, 0, 0, 0, 0);
 }
+
+TEST(AMap, canSetTowerObstacleAtomically)
+{
+    mock().expectOneCall("chMtxLock").withPointerParameter("lock", &map.lock);
```
antoinealb: Future idea: we could do something similar to std::lock_guard for chibios.

master-firmware/tests/test_strategy.cpp
```diff
@@ -176,3 +185,39 @@ TEST(Strategy, CanBuildTower)
     CHECK_TRUE(len > 0);
     CHECK_TRUE(goal.is_reached(state));
 }
+
+TEST(Strategy, CanBuildSecondTower)
+{
+    const int max_path_len = 10;
```
antoinealb: [nitpick] Those tests (CanDoWhatever) have a lot in common. If it was for something longer-lived than the strategy I would suggest refactoring them.

master-firmware/tests/test_strategy_helpers.cpp
```diff
@@ -12,25 +12,94 @@ TEST_GROUP(ABlockPositionComputer)
...
+TEST_GROUP(ABlockPositionComputer)
+{
+    se2_t poses[4] = {
+        se2_create_xya(850 - 160, 540 - 160, -45),
+        se2_create_xya(850 + 160, 540 - 160,  45),
+        se2_create_xya(850 + 160, 540 + 160, 135),
+        se2_create_xya(850 - 160, 540 + 160, 225),
+    };
+
+    void SE2_EQUAL(se2_t lhs, se2_t rhs, double tolerance=0.01)
```
antoinealb: Idea: We could create a set of cpputest utilities for use with our custom datatypes. Summer project probably.

master-firmware/src/robot_helpers/math_helpers.c
SyrianSpock: Good catch :sweat_smile:

master-firmware/src/strategy/state.h
SyrianSpock: no, but at least it's all in the same place.

master-firmware/tests/test_map.cpp
SyrianSpock: using RAII?

antoinealb: ja

master-firmware/src/strategy.cpp
SyrianSpock: good point

PR #115, opened 2018-04-07
Integrate UWB beacon firmware

antoinealb:
This PR merges my work from my semester project and onward. I dont really expect a full review on it honestly, but feel free to do it if needed :)

We will probably only use it to communicate between the two robots this year though.


uwb-beacon-firmware/src/ekf.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+    /** Returns the Jacobian of the state update function. */
+    virtual Jacobian G(State state, Input input)
+    {
+        (void) input;
+        (void) state;
+        return Jacobian::Identity();
+    }
+
+    std::pair<State, Covariance> predict(State mu, Covariance sigma, Input u)
```
SyrianSpock: [trolling] Could have been a `operator()` override to make it a functor :p

uwb-beacon-firmware/src/state_estimation_thread.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+        if (topic == range_topic) {
+            // If we got a range, feed it to the estimator
+            range_msg_t msg;
+            anchor_position_msg_t *anchor_pos;
+            messagebus_topic_read(topic, &msg, sizeof(msg));
+
+            // Discard messages with range greater than 1km, as they are
+            // probably the result of an underflow if the tag is too close to
+            // the anchor.
```
SyrianSpock: lol wut?

uwb-beacon-firmware/src/uavcan/topics_publisher.cpp
```diff
@@ -0,0 +1,199 @@
...
+    chThdCreateStatic(thd_wa, sizeof(thd_wa),
+                      NORMALPRIO,
+                      topics_watcher, NULL);
+}
+
+void topics_publisher_spin(Node &node)
+{
+    (void) node;
+
+    if (chBSemWaitTimeout(&imu_topic_signaled, TIME_IMMEDIATE) == MSG_OK) {
```
SyrianSpock: Why not publish directly when the wait triggers (above)?

uwb-beacon-firmware/src/ekf.hpp
antoinealb: actually i did not know about operator() when I built it in cpp, the ekf.py module is implemented that way ;)

uwb-beacon-firmware/src/state_estimation_thread.cpp
antoinealb: basically if you put 2 tags within like 5cm of each other, when computing the distance you get an underflow with an ill-calibrated system. was just easier to filter them.

uwb-beacon-firmware/src/uavcan/topics_publisher.cpp
antoinealb: Oh thats because of the API mismatch between libuavcan and messagebus. Let me explain this delightful interaction.

1. Messagebus watchgroups are blocking. Making them different than waiting, like having some semantics for "if something changed since last time i asked" would be very hard.
2. libuavcan is not thread safe. Therefore you **must** run actions before spinning or in a timer.
3. If you block in a libuavcan spin or timer, then you block the whole UAVCAN thread, which is not happy about it.

Solution: A blocking thread dispatches to a bunch of semaphores, which are then read without blocking. Sucks, but less than giving up completely on msgbus.

antoinealb:
Ah Travis, you never cease to please me.

Will fix tomorrow.

[troll] If the build was hermetic, it would of course never happen that the build fails on CI if it works on dev machine.

uwb-beacon-firmware/src/uavcan/topics_publisher.cpp
SyrianSpock: Works for me

uwb-beacon-firmware/src/state_estimation_thread.cpp
SyrianSpock: Fair enough

SyrianSpock:
All systems nominal, you can merge :rocket:

PR #117, opened 2018-04-11
Use ball gun on both robots

SyrianSpock:
This PR implements support for the ball gun module, and integrates it in the strategy of both robots. The ball gun sequences need more tuning, but the groundwork is done.

### DIFF
--- a/master-firmware/src/config.c
+++ b/master-firmware/src/config.c
@@ -67,6 +67,12 @@ static struct {
         parameter_t retracted;
         parameter_t deployed;
     } servo;
+    struct {
+        parameter_namespace_t ns;
+        parameter_t arm;
+        parameter_t charge;
+        parameter_t fire;
+    } turbine;
 } ballgun;
 
 static parameter_namespace_t arms_config, arms_main_config, motor_offsets_config;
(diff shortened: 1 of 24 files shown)


master-firmware/src/config.c
```diff
@@ -60,6 +60,31 @@ static struct {
     } right, left;
 } lever;
 
+static struct {
```
antoinealb: [idea] This whole file could probably be generated from the config.yml

master-firmware/src/strategy.cpp
```diff
@@ -548,6 +549,142 @@ struct BuildTowerLevel : actions::BuildTowerLevel {
...
+void strat_collect_wastewater(enum strat_color_t color)
+{
+    ballgun_tidy(&main_ballgun);
+    ballgun_deploy(&main_ballgun);
+    strategy_wait_ms(1000);
+
+    ballgun_charge(&main_ballgun);
+
+    trajectory_a_rel(&robot.traj, MIRROR(color, -95));
+    trajectory_wait_for_end(TRAJ_FLAGS_SHORT_DISTANCE /*TRAJ_FLAGS_ROTATION*/);
```
antoinealb: Why not TRAJ_FLAGS_ROTATION?

master-firmware/src/strategy.cpp
```diff
@@ -548,6 +549,142 @@ struct BuildTowerLevel : actions::BuildTowerLevel {
     }
 };
 
+void strat_collect_wastewater(enum strat_color_t color)
+{
+    ballgun_tidy(&main_ballgun);
+    ballgun_deploy(&main_ballgun);
+    strategy_wait_ms(1000);
```
antoinealb: Thats a lot of delay. Is it really needed ?

master-firmware/src/strategy.cpp
```diff
@@ -548,6 +549,142 @@ struct BuildTowerLevel : actions::BuildTowerLevel {
...
+        NOTICE("Filling water tower from %d %d", x_mm, y_mm);
+
+        if (!strategy_goto_avoid(MIRROR_X(m_color, x_mm), y_mm, MIRROR_A(m_color, -90), TRAJ_FLAGS_ALL)) {
+            return false;
+        }
+
+        strat_fill_watertower();
+
+        state.ballgun_state = BallgunState::IS_EMPTY;
+        state.balls_in_watertower += 8;
```
antoinealb: Full of hope hmm ?

master-firmware/src/strategy.cpp
antoinealb: Also I think there should be some error handling in there. Like if you are blocked you should maybe move away from the dispenser then try to retract. Otherwise you might break something.

master-firmware/src/strategy.cpp
SyrianSpock: Not tuned really

PR #123, opened 2018-04-15
Separate MessagePack config for Order and Chaos

nuft:
With this PR I propose to generate two different `config_msgpack.c` from the yaml configs of the two robots. This avoids having to `rm src/config_msgpack.c` when changing the build target.

It also fixes some annoying make warnings.

### DIFF
--- a/master-firmware/Makefile
+++ b/master-firmware/Makefile
@@ -55,8 +55,10 @@ endif
 
 ifeq ($(ROBOT), order)
   CONFIG_YAML = ../config_order.yaml
+  CONFIG_MSGPACK_C = src/config_msgpack_order.c
 else
   CONFIG_YAML = ../config_chaos.yaml
+  CONFIG_MSGPACK_C = src/config_msgpack_chaos.c
 endif
 
 #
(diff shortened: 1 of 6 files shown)


master-firmware/Makefile
```diff
@@ -285,29 +285,25 @@ endif
...
 
-clean:
-	@echo Cleaning
-	-rm src/config_msgpack.c
-	-rm -fR .dep $(BUILDDIR)
-	@echo
-	@echo Done
+CLEAN_RULE_HOOK:
+	@$(COLOR_PRINTF) "Remove MessagePack config"
+	-rm -f $(CONFIG_MSGPACK_C)
```
antoinealb: Maybe here remove all config_*.c ?

SyrianSpock:
+1 for having both configs and selecting based on a jumper on the shield. Then we just add a log telling us which robot has booted (and move gui init before color selection) so we can check that we flashed correctly via the screen easily.

nuft:
I implemented selectable config load, as suggested.
With this change we build the same binary for both robots and the loaded config is selected by the ROBOT_SELECT jumper on the shield.

nuft:
not yet, will merge when tested

nuft:
Tested on hardware, works well.

PR #126, opened 2018-04-15
Starter debounce for shaky hands

nuft:
...to avoid further problems during robot preparation :wink:


nuft:
Works on hardware... but now starts bothering me during setup.
How about passing the STARTER_MIN_ARMED_TIME as argument?

antoinealb:
Maybe press the color selection switch again for setup ?

SyrianSpock:
Either switch to color selection button for setup (but then do you need to press the same color?) or make the threshold an argument of the function. Both solutions are OK for me.

PR #128, opened 2018-04-23
Enable obstacle avoidance in polar mode

### LINKED ISSUE #121: Fix opponent beacon in non-XY trajectory mode
Currently the beacon handling code assumes (incorrectly) that we are always moving in XY mode, but angle_* and d_* mode are very useful and we should handle them.

See: `trajectory_helpers.c`.

SyrianSpock:
This PR fixes #121 partially by predicting collision with opponent when robot is moving in polar mode (distance or angle setpoints).

### DIFF
--- a/master-firmware/src/commands.c
+++ b/master-firmware/src/commands.c
@@ -417,6 +417,11 @@ static void cmd_traj_forward(BaseSequentialStream *chp, int argc, char *argv[])
         int32_t distance;
         distance = atoi(argv[0]);
         trajectory_d_rel(&robot.traj, distance);
+        int end_reason = trajectory_wait_for_end(
+            TRAJ_END_GOAL_REACHED | TRAJ_END_COLLISION |
+            TRAJ_END_OPPONENT_NEAR | TRAJ_END_ALLY_NEAR);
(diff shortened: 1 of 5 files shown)


master-firmware/src/commands.c
```diff
@@ -417,6 +417,11 @@ static void cmd_traj_forward(BaseSequentialStream *chp, int argc, char *argv[])
         int32_t distance;
         distance = atoi(argv[0]);
         trajectory_d_rel(&robot.traj, distance);
+        int end_reason = trajectory_wait_for_end(
+            TRAJ_END_GOAL_REACHED | TRAJ_END_COLLISION |
+            TRAJ_END_OPPONENT_NEAR | TRAJ_END_ALLY_NEAR);
+        trajectory_hardstop(&robot.traj);
+        chprintf(chp, "End readon %d\r\n", end_reason);
```
antoinealb: typo readon->reason

SyrianSpock:
Yep tested and working on robot. 
I chose not to support line/circle/clothoid modes because we don't use them, so I cannot test that it's working.

PR #132, opened 2018-04-27
Fix stack usage of GOAP

antoinealb:
Imagine everything you can do now that you can pass big structs to GOAP state :) What a lovely idea!

Very little change to API surface but basically we avoir having huge structs on the stack everywhere.

Also pulls unrelated change for the state topic.

Tested on nucleo but i had to disable a lot of stuff so it will required testing again in the full robot.


antoinealb:
Guess I will need to step by step it then. So Sad.

PR #134, opened 2018-04-29
Double the towers, double the fun

SyrianSpock:
This PR enables Chaos to build two towers :smiling_imp: 
It's a bit of a hack, and I haven't found a satisfying way of doing it (without solving the struct size issue). The alternative is to store these parameters in the config, but I don't think they ought to be stored there. Improvement ideas are welcome.


antoinealb:
Looks good to me, but a lot of hacks. Meh, it does not really matter, as the lifespan of this code will be < 2 weeks.

PR #135, opened 2018-05-02
Lever: Detect if pickup was successful

SyrianSpock:
All in the title, integrated in the strategy. Let there be GOAP!


master-firmware/src/lever/lever.c
```diff
@@ -134,3 +134,12 @@ se2_t lever_deposit(lever_t* lever, se2_t robot_pose)
...
     return blocks_pose;
 }
+
+void lever_tidy(lever_t* lever)
+{
+    lever_lock(&lever->lock);
+    lever_pump_set(lever, LEVER_PUMP_OFF);
+    lever_unlock(&lever->lock);
+
+    lever_retract(lever);
```
antoinealb: This should be inside the locked section, I think.

master-firmware/src/strategy.cpp
```diff
@@ -386,6 +386,29 @@ void strat_push_the_bee_v2(point_t start, float bee_height, float forward_motion
     trajectory_wait_for_end(TRAJ_FLAGS_SHORT_DISTANCE);
 }
 
+bool strat_lever_is_full(enum lever_side_t lever_side)
+{
+    bool full = true;
+
+    if (lever_side == LEVER_SIDE_LEFT) {
+        messagebus_topic_t* topic = messagebus_find_topic_blocking(&bus, "/lever/left");
```
antoinealb: Probably safer to use `messagebus_find_topic` (nonblocking) and handle the case where the topic does not exist. Just in case someone disables something or changes a topic name or whatever.

Also, this function could definitely use some refactoring, but given that this is in strategy... who cares.

master-firmware/src/lever/lever.c
SyrianSpock: `lever_retract` already locks inside https://github.com/cvra/robot-software/blob/master/master-firmware/src/lever/lever.c#L68

master-firmware/src/strategy.cpp
SyrianSpock: Refactoring? Ain't nobody got time for that

master-firmware/src/lever/lever.c
antoinealb: i know. Still lever_tidy is not atomic in its current config, it could be interrupted and changed between unlock and retract.

SyrianSpock:
Well it depends how many blocks you allow it to pickup (see actions provided to GOAP). Currently they both have only two blocks, so if one fails, they just build a single tower.

SyrianSpock:
Works on robot

antoinealb:
Does it work with latest changes or does it still drop cubes in bad ways ?

SyrianSpock:
Fixed now

PR #140, opened 2018-06-04
First draft of TDMA protocol

antoinealb:
This pull request contains the first draft of the TMDA protocol for the UWB beacons. It is not intended to be merged as-is, but rather to serve as a discussion topic.

### DIFF
--- a/uwb-beacon-firmware/doc/protocol_v2.md
+++ b/uwb-beacon-firmware/doc/protocol_v2.md
@@ -1,0 +1,55 @@
+# Design document for UWB protocol v2
+
+## Requirements
+
+* We want to update each robots' positions at 20 Hz.
+    - Each distance measurement requires a handshake of 3 packets.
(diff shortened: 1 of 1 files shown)


nuft:
I quickly read through the draft and it sounds good to me.

One remark on the requirements though: It might be useful to support the fourth beacon in the center - at least for communication.

antoinealb:
True, I forgot about the 4th beacon. Might be useful indeed, although in the worst case we could run a cable from it to a neighbor beacon that would then transmit over the air.

SyrianSpock:
I'm fine with assuming only 3 fixed beacons are supported.

SyrianSpock:
Would be good to support 4 fixed beacons for the landminesfree competition

antoinealb:
4 beacon it is then. Still not satisfied with the way to send data packets though :(

antoinealb:
Most likely never will be done

PR #143, opened 2018-08-12
Import minesweepers

SyrianSpock:
This PR imports the minesweepers project in here. The goal being to centralize all of our codebase in a monorepo.

Main changes:
- Move master-firmware and arm kinematics doc to `eurobot` folder
- Update the readme & CI scripts
- Import minesweepers repo

Still not sure about why Travis is still failing though.

Following PRs will be:
- Merging work on UWB (although only support one moving tag) + 2D plotting tool for UWB tests on the field
- Merging work on metal detector using the motor control board


SyrianSpock:
Both options have advantages/disadvantages:
1. Yes, but it's orthogonal
2. Yes, but the master-firmware is in fact "eurobot-software"
3. It's tricky, but git log/blame still work if you specify `--follow`
4. Until it's fixed :stuck_out_tongue:  

Why I moved master-firmware:
1. it makes more sense to group competition specific code and avoid spreading it all over the place
2. it's too odd to have a competition folder with no code (arbitrary reason)
3. ideally I would like more splitting and hierarchy to make reuse and navigation easier

but I'm still not 100% convinced of any of both options (moving vs not moving).

antoinealb:
> It's tricky, but git log/blame still work if you specify --follow

Which does not fix the github web interface for example ;) I know --follow, it's not über great.

> Yes, but the master-firmware is in fact "eurobot-software"

Not really, quite a few components of the code are Eurobot-specific. And I would like to be shown that we are really going to do more than 1 non-eurobot event... until then, YAGNI ;)

SyrianSpock:
> Which does not fix the github web interface for example ;) I know --follow, it's not über great.

Still seems to work on github :see_no_evil: https://github.com/cvra/robot-software/blame/import-minesweepers/eurobot/master-firmware/Makefile

> Not really, quite a few components of the code are Eurobot-specific. And I would like to be shown that we are really going to do more than 1 non-eurobot event... until then, YAGNI ;)

Then you don't want a `eurobot` directory at all? Or do I still get to move `bluetooth.txt`, `openlog` and specific docs to there?

antoinealb:
They implemented --follow on blame but not on history, wtf: https://github.com/cvra/robot-software/commits/import-minesweepers/eurobot/master-firmware/Makefile

> Then you don't want a eurobot directory at all? Or do I still get to move bluetooth.txt, openlog and specific docs to there?

Good question. I dont have a really strong opinion on it, but more of a gut feeling. Need to think about it.

SyrianSpock:
Dropped the idea of moving `master-firmware` to the `eurobot` directory. I still put docs and config files (bluetooth and openlog) in there.

nuft:
If we continue to use our software stack for non-eurobot applications I'm all for moving code in topic specific subdirectories. Although, I agree with @antoinealb that at the moment `master-firmware` contains a lot of generic code that will be shared between different applications. Perhaps this can be a continuous process where if some module is used in another project it can be moved to `lib`?
However, this would not exclude moving `master-firmware` to `eurobot` first.

PR #158, opened 2018-09-26
Implement differential base motion control

SyrianSpock:
Rover can now move forward or rotate (choose one at a time). The servos are actuated to achieve that (and wired :tada:).

Note: I had to put the remote control thread to sleep or the servos started going crazy (not keeping their position and jittered motion).

### DIFF
--- a/drive-board-firmware/src/remote_control.c
+++ b/drive-board-firmware/src/remote_control.c
@@ -95,6 +95,7 @@ static THD_FUNCTION(rc_thread_main, arg)
 
             // NOTICE("%f,%f", linear_x, angular_z);
             base_set_speed(&rover_base, linear_x, angular_z);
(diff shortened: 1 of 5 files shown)


drive-board-firmware/src/remote_control.c
```diff
@@ -95,6 +95,7 @@ static THD_FUNCTION(rc_thread_main, arg)
 
             // NOTICE("%f,%f", linear_x, angular_z);
             base_set_speed(&rover_base, linear_x, angular_z);
+            chThdSleepMilliseconds(10);
```
nuft: This is weird, will investigate when there is time. until then, ship it!

SyrianSpock: :shipit:

PR #159, opened 2018-09-27
UWB position publisher over ROS

SyrianSpock:


minesweepers/uavcan_bridge/nodes/position_publisher.py
```diff
@@ -0,0 +1,57 @@
...
+    tf.transform.translation.z = z
+
+    return tf
+
+def main():
+    if len(sys.argv) < 3:
+        print("usage: position_publisher.py uavcan_dsdl_path can_interface")
+        return
+
+    dsdl_path, can_interface = sys.argv[1], sys.argv[2]
```
nuft: use rosparam for this?

SyrianSpock: yes, I still need to write a roslaunch and feed the values from rosparam. Stay tuned :tv:

Issue #144, opened 2018-08-12
Make PI metal detector using motor control board


SyrianSpock:
Given the time constraints of the competition, we need to reuse the motor board to detect metal

MVP1:
- [x] Generate pulse to power the coil
- [x] Read discharge signal from ADC
- [x] Send signal over UAVCAN
- [x] Fit the discharge signal with a delayed exponential
- [x] Validate the setup by experiment
- [x] Compensate for temperature drift

PR #160, opened 2018-09-27
Metal detector

SyrianSpock:
This PR adds:
- the firmware to turn a motor board into an EMI metal detector
- a tool in CVRA studio that fits the EMI signal to detect the presence of metal

The tool is a hack-ish and adds a dependency of CVRA studio on scipy. In a following PR I will remove this tool out and make ROS nodes of out it. It will most likely be
- one UAVCAN bridge node to forward the raw signals
- one processing node to fit the curve and classify the even). 

solves #144


metal-detector-firmware/src/uavcan/uavcan_node.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <uavcan_stm32/uavcan_stm32.hpp>
+#include <uavcan/protocol/SoftwareVersion.hpp>
+
+#include <version/version.h>
+
+#include <main.h>
+#include "uavcan_node.h"
+#include "parameter_server.hpp"
+
+#include <cvra/metal_detector/EMIRawSignal.hpp>
```
nuft: Compilation fails here:
```
Compiling uavcan_node.cpp
./src/uavcan/uavcan_node.cpp:15:10: fatal error: cvra/metal_detector/EMIRawSignal.hpp: No such file or directory
 #include <cvra/metal_detector/EMIRawSignal.hpp>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
Did you forget to add the custom UAVCAN datatypes?

SyrianSpock: Good catch. Yes I forgot to add the file while rebasing 😅 It's fixed now

PR #168, opened 2018-11-08
Add a tool to format changed files

antoinealb:
This tool can be used prior to committing to format all changed files in
the repository to make sure it respect the coding style.


antoinealb:
I know that we want to switch to `clang-format` but we can eventually start introducing the team to the usage of this tool. Then when we use clang-format, we simply add it in this script, keeping the same frontend.

PR #169, opened 2018-11-14
Store touchscreen config

Lu-ni:
Each touchscreen need to have his own configuration to work well.

* Store touch screen configuration in each robot config.

Next step will be to implement touch screen calibration.

### DIFF
--- a/master-firmware/src/config.c
+++ b/master-firmware/src/config.c
@@ -318,12 +318,12 @@ void config_init(void)
     parameter_scalar_declare_with_default(&main_arm.control.y.ilimit, &main_arm.control.y.ns, "i_limit", 0);
 
     parameter_namespace_declare(&screen.ns, &master_config, "screen");
-    parameter_scalar_declare_with_default(&screen.ax, &screen.ns, "ax", 0);
-    parameter_scalar_declare_with_default(&screen.ay, &screen.ns, "ay", 0);
-    parameter_scalar_declare_with_default(&screen.bx, &screen.ns, "bx", 0);
-    parameter_scalar_declare_with_default(&screen.by, &screen.ns, "by", 0);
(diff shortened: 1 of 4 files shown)


antoinealb:
Could you document somewhere how those values can be obtained?

master-firmware/src/config.c
```diff
@@ -306,6 +316,14 @@ void config_init(void)
     parameter_scalar_declare_with_default(&main_arm.control.y.ki, &main_arm.control.y.ns, "ki", 0);
     parameter_scalar_declare_with_default(&main_arm.control.y.kd, &main_arm.control.y.ns, "kd", 0);
     parameter_scalar_declare_with_default(&main_arm.control.y.ilimit, &main_arm.control.y.ns, "i_limit", 0);
+
+    parameter_namespace_declare(&screen.ns, &master_config, "screen");
+    parameter_scalar_declare_with_default(&screen.ax, &screen.ns, "ax", 0);
+    parameter_scalar_declare_with_default(&screen.ay, &screen.ns, "ay", 0);
+    parameter_scalar_declare_with_default(&screen.bx, &screen.ns, "bx", 0);
+    parameter_scalar_declare_with_default(&screen.by, &screen.ns, "by", 0);
+    parameter_scalar_declare_with_default(&screen.cx, &screen.ns, "cx", 0);
```
antoinealb: Might be better not to have any defaults as the default value of 0. will cause the screen to act in weird ways. I would prefer to have an assert fail.

Lu-ni:
Tested and working now

Lu-ni:
> Could you document somewhere how those values can be obtained?

Where should I put the file ? A README.md in the gui folder (../src/gui/) ?

antoinealb:
Yes I think this would work.

antoinealb:
Yes I think this would work.

PR #171, opened 2018-11-17
Initial implementation of Raft

antoinealb:
This PR implements a first working version of the raft consensus protocol. Please see the included paper for details on how Raft works.

Things that are complete so far:

* Leader election
* Log replication
* Commit index

Things that are not yet implemented, but should come in subsequent PR:

* Forwarding of committed log entries to the state machine
* Log compaction

Things that will not be included:

* Online cluster membership change, because it is safe to assume that the cluster will stay constant during an entire game.

I wanted to open this PR as soon as possible, even if it is not complete (many todos in the code) in order to have a smaller PR size later on.


lib/raft/raft.hpp
```diff
@@ -0,0 +1,512 @@
...
+            bool vote_granted;
+        } vote_reply;
+
+        struct {
+            int count;
+            Index leader_commit;
+            Term previous_entry_term;
+            Index previous_entry_index;
+            // TODO: How many entries
+            LogEntry<typename StateMachine::Operation> entries[10];
```
SyrianSpock: Consider wrapping `10` in a const variable at the top of the file together with the TODO note

lib/raft/raft.hpp
```diff
@@ -0,0 +1,512 @@
...
+                reply.type = Message::Type::VoteReply;
+                reply.vote_reply.vote_granted = false;
+
+                const auto req_lli = msg.vote_request.last_log_index;
+                const auto req_llt = msg.vote_request.last_log_term;
+
+                if ((msg.term > term && log.last_index() <= req_lli && log.last_term() <= req_llt)
+                    ||
+                    (msg.term == term &&
+                     msg.from_id == voted_for)) {
```
SyrianSpock: Consider grouping logic expressions in meaningful variables to improve readability, for example:
```cpp
const bool best_candidate = msg.term > term && log.last_index() <= req_lli && log.last_term() <= req_llt;
const bool elected_candidate = msg.term == term && msg.from_id == voted_for;

if (best_candidate || elected_candidate) {
```

lib/raft/raft.hpp
```diff
@@ -0,0 +1,512 @@
...
+
+            case Message::Type::VoteReply: {
+                DEBUG("Got a VoteReply(granted = %d)", msg.vote_reply.vote_granted);
+                // TODO: How about potential duplicate votes =
+
+                if (node_state != NodeState::Candidate) {
+                    break;
+                }
+
+                vote_count ++;
```
SyrianSpock: Shouldn't you check first `if (msg.vote_reply.vote_granted)` before counting this vote?

lib/raft/raft.hpp
```diff
@@ -0,0 +1,512 @@
...
+
+                if (msg.vote_reply.vote_granted) {
+                    if (2 * vote_count >= peer_count) {
```
SyrianSpock: Should this be absolute majority instead (`2 * vote_count > peer_count`)?

lib/raft/raft.hpp
```diff
@@ -0,0 +1,512 @@
...
+                    }
+                }
+                break;
+            }
+        }
+
+        return false;
+    }
+
+    void start_election()
```
SyrianSpock: Since all members are public, you could consider making these free functions and inject the class as argument

lib/raft/raft.hpp
```diff
@@ -0,0 +1,512 @@
...
+};
+
+template <typename StateMachine>
+class State {
+public:
+    using Message = raft::Message<StateMachine>;
+    using Peer = raft::Peer<StateMachine>;
+    NodeId id;
+    Peer **peers;
+    int peer_count;
```
SyrianSpock: No `std::vector<std::shared_ptr<Peer>>` so sad :crying_cat_face:

lib/raft/tests/messages_comparator.hpp
```diff
@@ -0,0 +1,16 @@
+#include <CppUTest/TestHarness.h>
+#include <CppUTestExt/MockSupport.h>
+#include "../raft.hpp"
+#include "test_state_machine.hpp"
+
+class RaftMessageComparator : public MockNamedValueComparator
+{
+public:
+    virtual bool isEqual(const void* object1, const void* object2);
+    virtual SimpleString valueToString(const void* object);
```
SyrianSpock: Consider implementing `operator==` and `operator<<` for `Message` if that makes sense

antoinealb:
fixed, rebased and formatted.

PR #173, opened 2018-11-29
Move compiler defines to header file

Lu-ni:
It now store the compiler define in:
* master-firmware/shell.config 
* master-firmware/uavcan.config
* master-firmware/lwip.config


master-firmware/shell.include
```diff
@@ -0,0 +1 @@
+#define SHELL_CONFIG_FILE
```
nuft: The shell is part of ChibiOS, so I would name this file `chibios.include`

Issue #118, opened 2018-04-12
generate config.c from yaml


antoinealb:
So it appears to me that the config.c file from master firmware has a lot of boilerplate code that can probably be autogenerated from config.yml anyways. So we should do that.

I think this can be a good beginner project for someone interested in build tools (nobody is interested in build tools?)

PR #174, opened 2018-12-03
Generate config boilerplate

SyrianSpock:
This PR tackles #118 by generating the C boilerplate (struct + init) from one or multiple YAML files.
The Python code is unit tested (TDD for the win), we can consider running Python tests in the CI in the near future.

I tested it on one robot and everything seems to work.


nuft:
LGTM

However, I can not help but notice that the `config_to_c.py` and `config_to_msgpack.py` are somewhat redundant which add more complexity to the project. There is a lot of code involved going from yaml -> msgpack -> config tree -> parameter object.

Perhaps we should start to rethink how we handle the config tree.
Here are some random thoughts:
- Our config tree is known at compile time but needs code generation because the parameter subsystem builds the tree at runtime. Maybe the linker actually does a better job finding the parameter objects, which would eliminate all the boilerplate code for `parameter_scalar_declare()` and `parameter_find()`.
- What the parameter subsystem actually provides is a runtime "map" from names to parameter objects with thread-safe access and synchronization methods.
- This functionality is very similar to messagebus. There is potential of unifying the two subsystems.

antoinealb:
> Perhaps we should start to rethink how we handle the config tree.

Agree, what we are currently doing seems... weird.

> Our config tree is known at compile time but needs code generation because the parameter subsystem builds the tree at runtime. Maybe the linker actually does a better job finding the parameter objects, which would eliminate all the boilerplate code for parameter_scalar_declare() and parameter_find().

I am not completely sure how the linker could find it.
We often have functions that declare a set of parameter inside a namespace, for example to generate the 4 parameters needed for a PID controller. The linker could not reuse this.
Also, I feel that most of the boilerplate comes from fetching the parameter value, not from declaring it.

> What the parameter subsystem actually provides is a runtime "map" from names to parameter objects with thread-safe access and synchronization methods.
> This functionality is very similar to messagebus. There is potential of unifying the two subsystems.

While they are similar in terms of thread safety objectives, I feel that it is better to keep them separate from now.
Their use case is very different (control path vs data path), and I am not sure if we can provide a good API for both.
Plus, they are not a lot of code, so keeping them separate is not a huge burden (esp. since at least `parameter` is pretty stable now).

An idea that I had to make the parameter order easier to use would be to do so using C++ statics, which would look like this:


```cpp
Namespace root("/");
Namespace mypid("/mypid");
Parameter<float> kp("/mypid/kp", 1.);

void pid_thread(void *p)
{
    while (true) {
        if (mypid.changed()) {
            pid_set_kp(kp); // Parameter<float> implements operator float();
        }
    }
}

// somewhere before, initialise the parameter tree from the statics
```

It reduces the amount of boilerplate without relying on "magic" from linker or build tools.
What do you think ?

SyrianSpock:
> However, I can not help but notice that the `config_to_c.py` and `config_to_msgpack.py` are somewhat redundant which add more complexity to the project. There is a lot of code involved going from yaml -> msgpack -> config tree -> parameter object.

I saw that, if you have time refactor it, I'd be happy to review the PR.

> Their use case is very different (control path vs data path), and I am not sure if we can provide a good API for both.
Plus, they are not a lot of code, so keeping them separate is not a huge burden (esp. since at least `parameter` is pretty stable now).

Agree. Use before reuse.

> It reduces the amount of boilerplate without relying on "magic" from linker or build tools.

Indeed we can use C++ to make it syntactically better. We should rethink the parameter tree though, consider merging parameter and parameter_flash_storage in the same library and provide more (serialization/introspection, C++ interface).

PR #178, opened 2018-12-16
Explicit ownership of obstacle_avoidance

### LINKED ISSUE #137: Cleanup obstacle avoidance module
- [ ] Inject a struct pointer to all `oa` function calls
- [ ] Explicitly handle ownership of `oa` in the map server thread
- [ ] All queries to `oa` should be properly synchronized.

antoinealb:
Previously, obstacl_avoidance used to be a global singleton, leading to
all sorts of issues with shared ownership, proper locking and so on.

This commit changes the API so that all calls to the obstacle avoidance
module take a pointer to the data to use. This also meant changing a few
tests.

On the robot the obstacle avoidance data is now handled by the map
server. The only way to access it is through an API that ensures it is
properly locked.

Fixes #137


master-firmware/src/base/map_server.cpp
```diff
@@ -118,3 +124,15 @@ void map_server_start(enum strat_color_t color)
...
+{
+    chMtxLock(&map_data.map_lock);
+    return &map_data.map;
+}
+
+void map_server_map_release(struct _map* map_ptr)
+{
+    osalDbgAssert(map_ptr == &map_data.map, "wrong map ptr");
+    chMtxUnlock(&map_data.map_lock);
+}
```
SyrianSpock: It's safer if you do the same with RAII, after all this is a `cpp` file :angel: 
```cpp
struct MapData
{
    MapData(void)
    {
        chMtxLock(&map_data.map_lock);
    }
    ~MapData(void)
    {
        chMtxUnlock(&map_data.map_lock);
    }
    operator struct _map*() const 
    {
        return &map_data.map;
    }
}
```

antoinealb: yep yep, but those functions are called from C files... Otherwise of course it would have been RAII all the way :p

antoinealb:
Should we test this on hardware before merging ? Do we have a working setup where this could be tested ?

master-firmware/src/base/map_server.cpp
SyrianSpock: time to make to `git mv strategy_helpers.c strategy_helpers.cpp`?

SyrianSpock:
> Should we test this on hardware before merging ?

Yes, it's safer. I'll test it and merge.

> Do we have a working setup where this could be tested ?

That's our secret, we ~always~ almost always have a working setup

SyrianSpock:
Tested on robot, and all looks good for this refactoring :+1:

PR #181, opened 2018-12-28
GUI simulator + new menu system

antoinealb:
I don't know if this is worthy of being merged yet. But here is my implementation of a menu system written in c++, which makes it easier to write new pages. Included with it is a simulator for menus that works on macOS (not tested on Linux).


SyrianSpock:
Had to fix the cmake to build on Linux, works now.
We can merge it now IMO, I don't see any blockers except fixing the build by committing the missing `gui/pages/PositionPage.h` :wink:.

Issue #136, opened 2018-05-06
Studio: Improve setpoint publisher widget


SyrianSpock:
The current setpoint publisher (part of `cvra_studio pid_plot`) was rushed, so it's missing some features:
- [x] The node is addressed by ID, should be by name through a list of available nodes
- [x] The setpoint should be either symmetric or not (alternate value and 0), currently it's symmetrically alternated only (positive/negative)
- [x] Dump config to yaml for easier PID tuning

PR #182, opened 2019-01-19
Export config from PID plotter

Lu-ni:
Print the YAML config in the terminal.

ex:
```YAML
    demo:
        control:
           mode: 0
            current:
                i_limit: 42.0
                kd: 12.0
                ki: 42.0
                kp: 42.0
            position:
                i_limit: inf
                kd: 32.0
                ki: 3.0
                kp: 0.0
            velocity:
                i_limit: inf
                kd: 0.0
                ki: 0.0
                kp: 0.0
           torque_limit: inf
           velocity_limit: 3.4028234663852886e+38
           acceleration_limit: inf
           low_batt_th: -1.0000303983688354
        motor:
           torque_cst: 1.0
        encoders:
            primary:
                p: 2
                q: 1
                ticks_per_rev: 1024
            secondary:
                p: 3
                q: 1
                ticks_per_rev: 1024
        potentiometer:
           gain: 1.0
           zero: 0.0
        rpm:
           phase: 0.0
        thermal:
           Cth: inf
           Rth: 1.0
           current_gain: 1.0
           max_temp: inf
        streams:
           current: 0.0
           velocity: 10.0
           position: 10.0
           index: 10.0
           enc_pos: 0.0
           motor_pos: 10.0
           motor_torque: 10.0
```

Ready to bake

### DIFF
--- a/tools/studio/cvra_studio/commands/param_tree.py
+++ b/tools/studio/cvra_studio/commands/param_tree.py
@@ -94,12 +94,6 @@ class ParameterTreeModel(NodeStatusMonitor):
                         for x,y in val.items():
                             print("\t\t{}: {}".format(x,y))
                     else: print("\t{}: {}".format(key,val))
-    #def myprint(self,d):
-    #    for k, v in d.items():
-    #        if isinstance(v, dict):
-    #            self.myprint(v)
-    #        else:
-    #            print("{0} : {1}".format(k, v))
 
     def _save_params_callback(self, event):
         if not event: self.logger.warning('Unable to save parameters')
--- a/tools/studio/cvra_studio/commands/param_tree.py
+++ b/tools/studio/cvra_studio/commands/param_tree.py
@@ -4,6 +4,7 @@ import queue
 import threading
 import time
 import sys
+import yaml
 
 import uavcan
 from PyQt5.QtGui import QFont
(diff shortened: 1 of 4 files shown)


antoinealb:
Note sure what this is for, but your print_param function does not correctly handle many nested levels. Use `yaml.dump` instead.

Example:

```python
import yaml

params = {
    'pid': {
        'kp': 12.,
        'ki': 8,
        'kd': 42,
        'foo': 'bar',
        'baz': {
            'lol': 21
        }
    }
}

print(yaml.dump(params, default_flow_style=False))
```

SyrianSpock:
This is the last improvement listed in #136

SyrianSpock:
@antoinealb won't work because `params` are stored as a `NestedDict`, a custom class I defined as datamodel that makes it easier to deal with hierarchies. @Lu-ni would need to implement a `NestedDict` to `dict` converter to use PyYaml :cry:

SyrianSpock:
> @antoinealb won't work because params are stored as a NestedDict, a custom class I defined as datamodel that makes it easier to deal with hierarchies. @Lu-ni would need to implement a NestedDict to dict converter to use PyYaml cry

which it turns out is straightforward. `NestedDict` derives from `defaultdict` (from collections) that converts to dict :tada:

PR #183, opened 2019-01-20
Publish sensor-board color reading 

nuft:


uavcan_data_types/cvra/sensor/20141.ColorTCS3472.uavcan
```diff
@@ -0,0 +1,8 @@
+#
```
SyrianSpock: why not call it simply `20141.Color.uavcan`?

nuft: Right, this message is quite generic. Originally, I wanted to add sensor specific configuration parameters, which became too complicated.
How about `20141.ColorRGBC.uavcan` or `20141.ColorRGBC16.uavcan`?

SyrianSpock: `20141.ColorRGBC.uavcan` sounds good to me

nuft: done

PR #184, opened 2019-01-24
Proximity beacon velocity control

Lu-ni:
#### PID plotter
* clear the plot when you change topic in the comboBox
* Fix crash when setting a parameter twice 
Wait before merging. I want to do an improvement before
#### proximity beacon
* Listen to setpoints over UAVCAN
* Load config from flash on boot


antoinealb:
So no merge ?

Lu-ni:
Not yet, I was a bit rushy on this one.
I'll reopen it soon :)

PR #188, opened 2019-02-26
Port IP UAVCAN bridge to Linux

antoinealb:
Using advanced wizardry, I was able to craft the incantations to open a network tap device on the penguin OS. It should now be able to participate in the UWB network.

@SyrianSpock can you test and report please ? Or we can test on thursday, i'll be there.


SyrianSpock:
forgot to pull it before going to the bunker, let's test it on Thursday

Issue #57, opened 2017-09-14
Use chibios-syscalls instead of the default syscalls provided by ChibiOS


SyrianSpock:
We only use chibios-syscalls in the can io firmware, others use the default syscalls from ChibiOS.

This should be more uniform.

SyrianSpock:
fixed in #192

PR #192, opened 2019-03-09
Unify C++ stubs

### LINKED ISSUE #191: UWB beacon firmware is built with exceptions enabled
I noticed that for the uwb beacon all the exception handling code is linked with the firmware.

We should investigate and try to remove the cause as it blows up the binary size and makes debugging harder.

SyrianSpock:
Moved C++ stubs to lib/chibios-syscalls to address #57. I observed that for the UWB firmware the bin is 40kB smaller (236556 to 194972 bytes). May fix #191


lib/chibios-syscalls/cpp_stubs.cpp
```diff
@@ -10,55 +11,33 @@ int __cxa_guard_acquire(__guard*);
...
 void __cxa_pure_virtual();
-void* __dso_handle = NULL;
+void* __attribute__((weak)) __dso_handle = NULL;
 void _exit(int status);
 pid_t _getpid(void);
 int _kill(int pid, int sig);
 }
 
-void __cxa_pure_virtual()
+void __attribute__((weak)) __cxa_pure_virtual()
```
antoinealb: Why make them as weak linking ? I think the libc++ version is already weak, so what is the point ?

SyrianSpock: was still looking at that. Our build is messier than I thought, so it took a while to fix it for all boards. Let's see if Travis is happy :crossed_fingers:

SyrianSpock:
Tested and working on master, can-io, sensor, and proximity beacon boards.

lib/chibios-syscalls/malloc_lock.c
```diff
@@ -1,20 +1,35 @@
 #include <ch.h>
+#include <hal.h>
+
 #include <malloc.h>
 
-static MUTEX_DECL(lock);
```
nuft: Why remove the static mutex initializer? (MUTEX_DECL() initializes the lock already)
This adds complexity because it requires calling `malloc_lock_init()` instead of letting the data section initializer do the job.
It also adds the overhead of `if (lock_initialized)` check.

antoinealb: Agree, MUTEX_DECL is the way to go. I guess its because this implementation was done differently in two boards and salah copied it over ?

lib/chibios-syscalls/dynamic_alloc_stubs.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <hal.h>
+
+#include <stdio.h>
+#include <errno.h>
+
+#ifndef CVRA_NO_DYNAMIC_ALLOCATION
+# define CVRA_NO_DYNAMIC_ALLOCATION 0
+#endif
+
+#if CVRA_NO_DYNAMIC_ALLOCATION
```
nuft: The default implementation of `operator new()` can throw a `bad_alloc` exception, which we don't want.
I think until now we never had code that called `new()`, so it's not a problem at the moment to leave the default implementation.
So, just to keep in mind...

lib/chibios-syscalls/malloc_lock.c
nuft: Just to be clear, this relies on the data section being initialized before any code that can call malloc (such as static constructors). But this is the case anyway, so no problem.

> I guess its because this implementation was done differently in two boards

Yes, it was done this way in master-firmware.

SyrianSpock: > I guess its because this implementation was done differently in two boards and salah copied it over ?

:see_no_evil: yes, of course I had to trust the master-firmware implementation.
I need to test on hardware again :crying_cat_face:

SyrianSpock:
all boards tested and operational, ready to merge if there are no more change requests :rocket:

PR #197, opened 2019-03-17
WIP: Reenable on-board SDCard for logging

nuft:
We disabled SD card logging because it didn't work during initial tests.
Probably this was due to missing pull-ups on the SD D0-4 and CMD lines.
I tested it on one board and it works reliably, yay 👻 

However, it only works if `gui_start()` is commented out. Otherwise it panics with a UsageFault 🔥   
The stack backtrace looks weird, it claims the error is at address 0x00000000 😱

My guess is that the fatfs routines in `log_message()` use too much stack and things start to break.
Need to investigate more...

### DIFF
--- a/master-firmware/src/board/board.h
+++ b/master-firmware/src/board/board.h
@@ -462,10 +462,10 @@
                                      PIN_PUPDR_FLOATING(GPIOC_ETH_RMII_RXD1) | \
                                      PIN_PUPDR_PULLUP(GPIOC_ENCODER_RIGHT_CHA) | \
                                      PIN_PUPDR_PULLUP(GPIOC_ENCODER_RIGHT_CHB) | \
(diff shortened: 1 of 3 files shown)


antoinealb:
Good job on finding the cause! I agree, sounds likely that the issue is a stack overflow somehow, will take a look at it later.

antoinealb:
Rebased and added detection for recursive logging. Will investigate once I get some time to play with a robot.

SyrianSpock:
Any chance SD card logging is back for this year's competitions?

nuft:
While I'm not sure if we're going to use logging extensively, I think we should definitely merge this.
I can do it as soon as I get the chance to test it on hardware.

nuft:
Obsolete, changed hardware

PR #199, opened 2019-04-01
add remote script for all bootloader_*

Lu-ni:
This PR add a script to launch remotely (via the raspi) the following cmd : 
 * bootloader_change_id
 * bootloader_read_config
 * bootloader_write_config
 * bootloader_flash
 * bootloader_run_app

### DIFF
--- a/tools/raspi_debug_server/README.md
+++ b/tools/raspi_debug_server/README.md
@@ -26,6 +26,10 @@ Get chibios shell
 ```BASH
 ./shell-ssh.sh
 ```
+Copy your ssh id for autologing
+```BASH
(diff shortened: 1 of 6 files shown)


Lu-ni:
> Not a huge fan of having ssh-copy-id done automatically for me, especially if I don't expect it.

I removed it completely. It would be boring to be asked each time if we want ssh-copy-id. 

> Also add `set -eu -o pipefail` at the beginning of your bash scripts. See https://stackoverflow.com/questions/13468481/when-to-use-set-e for explanations.

Thank you for the tips ! :)

PR #200, opened 2019-04-02
Arm control system design

SyrianSpock:
This PR proposes a design for the control system of the arm. Heavily inspired by Mikrob's Aversive control system architecture. 
I also implemented the lower level (system / state estimator / controller) on:
- a toy problem (1D pendulum) for TDD
- a concrete problem: joint-space control of the 3DOF arm for this year

I need feedback! (pun intended)


lib/golem/system.h
```diff
@@ -0,0 +1,79 @@
...
+public:
+    System() = default;
+    System(const System&) = delete;
+    System& operator=(const System&) = delete;
+    System(System&&) = default;
+    System& operator=(System&&) = default;
+
+    template <typename T>
+    System(T&& impl)
+        : m_impl(new model_t<typename std::decay<T>::type>(std::forward<T>(impl)))
```
antoinealb: meesa no like malloc calls()... Couldn't we just hold a reference to some external object ?

lib/golem/tests/pendulum.h
```diff
@@ -0,0 +1,82 @@
...
+        update(0);
+    }
+
+    Position2D get() const
+    {
+        return pos;
+    }
+    void update(const float& angle)
+    {
+        pos.x = LENGTH * std::cos(angle);
```
antoinealb: Please use `std::cosf` instead of `std::cos` since we only have single precision FPU.

master-firmware/src/manipulator/hw.cpp
```diff
@@ -0,0 +1,39 @@
...
+    std::array<float, 3> angles;
+    for (size_t i = 0; i < 3; i++) {
+        angles[i] = directions[i] * read_motor_position(motors[i]) - offsets[i];
+    }
+    return angles;
+}
+
+void System::apply(const std::array<float, 3>& angles)
+{
+    for (size_t i = 0; i < 3; i++) {
```
antoinealb: nitpick: `for (auto i = 0u; i < 3; i++)`

lib/golem/doc/system-design.md
```diff
@@ -0,0 +1,80 @@
...
+    * Avoid obstacles
+    * Limit maximum velocity and torque on the joints
+- Detect when the arm is blocked
+
+For that, we'll get inspiration from [Mikrob technology's control system design in aversive](https://wiki.droids-corp.org/articles/a/v/e/Aversive/Asservissement_Microb_2008.html).
+It provides us with a modular clean architecture that separates concerns properly.
+The separate components can be composed to implement complex behavior, and some components are optional.
+Here follows a diagram of the proposed design.
+
+![Diagram of the proposed arm control system design](figures/system.pdf)
```
antoinealb: maybe commit it as .png instead, so that it can be viewed on github directly

lib/golem/doc/system-design.md
```diff
@@ -0,0 +1,80 @@
...
+
+Each of the main blocks (motors, control loop, trajectory, and blocking detection) run in their own thread.
```
antoinealb: How do they communicate ?

lib/golem/doc/system-design.md
```diff
@@ -0,0 +1,80 @@
...
+
+The state estimator is responsible for gathering sensor measurements (i.e. joint positions, velocities, and torques) and estimate the state of the arm.
+The simplest possible state estimator only knows about the arm's kinematics and dynamics, such that it can convert from joint space to cartesian space, and viceversa.
+A more elaborate state estimator can be a Kalman filter which attempts to model the presence of noise in the input joint measurements and filter it out.
+
+### Consign filter
+
+The consign filter is an optional block.
+It takes a consign (desired target arm state) and smoothes it, usually in order to make it feasible.
+It can be a limited acceleration filter (ramp) or a limited velocity filter (quadramp).
```
antoinealb: If I am not mistaken, at least in Aversive the ramp is limited velocity, while quadramp limits accel + speed.

antoinealb: Also, how do you handle joint space limits such as torque limits ?

lib/golem/doc/system-design.md
```diff
@@ -0,0 +1,80 @@
...
+### Trajectory manager
+
+The trajectory manager receives target positions and sends them to the control loop.
+Then, it regularly polls the arm state to check if the target was reached (given a goal tolerance) to stop the control loop.
+This prevents the arm from shaking when near its end goal.
+
+### Path planner
+
+The path planner is aware of the obstacles around the arm.
+Thus, given a target position, it can compute the path circumventing obstacles and break it down into intermediate positions.
```
antoinealb: So a path planner outputs a discrete series of waypoint (like for navigation) or a continuous path ?

lib/golem/state_estimator.h
```diff
@@ -0,0 +1,79 @@
...
+    template <typename T>
+    StateEstimator(T&& impl)
+        : m_impl(new model_t<typename std::decay<T>::type>(std::forward<T>(impl)))
+    {
+    }
+
+    template <typename T>
+    StateEstimator& operator=(T&& impl)
+    {
+        m_impl.reset(new model_t<typename std::decay<T>::type>(std::forward<T>(impl)));
```
antoinealb: Same as before, not sure if I like malloc here.

master-firmware/src/priorities.h
```diff
@@ -17,6 +17,7 @@
 #define MAP_SERVER_PRIO (NORMALPRIO + 1)
 #define COLOR_SEQUENCE_SERVER_PRIO (NORMALPRIO)
 #define ARMS_CONTROLLER_PRIO (NORMALPRIO)
+#define MANIPULATOR_THREAD_PRIO (NORMALPRIO)
```
antoinealb: Reading this file makes me think "Those guys have no idea how to allocate priorities" :D

lib/golem/system.h
SyrianSpock: I'm thinking to use CRTP instead of Sean Parent's runtime polymorphism. Trying to avoid inheritance at all costs :angel:

lib/golem/tests/pendulum.h
SyrianSpock: good catch

- [ ] use cosf

master-firmware/src/manipulator/hw.cpp
SyrianSpock: I don't like to abuse `auto`

lib/golem/doc/system-design.md
SyrianSpock: - [ ] change drawing to png file

lib/golem/doc/system-design.md
SyrianSpock: Magic! Through function calls. For example you can set the controller's target via a function call from another thread. Still need to think about synchronization in the implementation :thinking:

lib/golem/doc/system-design.md
SyrianSpock: I'd say that's for the controller to manage (e.g. MPC). The consign filter may be discarded in some cases.

lib/golem/doc/system-design.md
SyrianSpock: Discrete series of waypoints.

- [ ] clarify that path planner outputs a discrete series of waypoints

master-firmware/src/priorities.h
SyrianSpock: i-dont-know-what-im-doing

SyrianSpock:
> So I understand it was tested on hardware ? Good performance ?

Yes, the code runs fine but I can't test the control approach, I still need to tune the motors' PIDs :crying_cat_face:, but will do ASAP.

> Also I wonder if it would be good to pass around parameter namespaces so that the various modules knew where they should register their params.

Also thinking about that, but my code started getting big and I don't like long-lived branches. Will see how to add that afterwards :thinking:

lib/golem/system.h
SyrianSpock: - [ ] get rid of malloc

antoinealb: Hmmm I think here inheritance is actually A-ok, isn't it ? And at least it will not blow in our face in strange ways ;)

master-firmware/src/manipulator/hw.cpp
antoinealb: But, but... "effective modern c++"

lib/golem/system.h
SyrianSpock: it's even less code with CRTP :heart_eyes_cat:

lib/golem/state_estimator.h
SyrianSpock: bye bye malloc, hello static polymorphism

SyrianSpock:
:heavy_check_mark: PIDs tuned for all motors
:heavy_check_mark: Inverse kinematics controller validated on prototype arm

Ready to merge :rocket: 

I'll follow up with control target filtering (to smooth the motion), implement the trajectory management and clean up the code to remove the old arm controller.

PR #201, opened 2019-04-16
[WIP] Integration of the new arm

SyrianSpock:
The first instance of the new arm was assembled and wired, now is time to make it move, with magic... I mean software 🧙. I had to change a few things on the motor board's firmware which worry me a bit, I need a second pair of eyes to have a look (also it's late).

Changes:
- Motor board: changed the way we compute the current constant
- Add a script to get motor config from datasheet values
- Configure motors of the new arm, and tune the current PID loops (waiting for a new 3D printed tool to block the motors better and improve the controller gains).

Note: I also discovered that we store the transmission fraction parts (p and q) on the motor board as `uint16_t`, so we quickly overflow. I need to see if we can refactor to use `uint32_t` easily, so we don't have to round the transmission factor :sweat_smile: .


motor-control-firmware/src/control.c
```diff
@@ -369,7 +370,7 @@ static void update_parameters(void)
     if (parameter_namespace_contains_changed(&motor_params.ns)) {
         if (parameter_changed(&motor_params.torque_cst)) {
             float transmission = (float)control_feedback.primary_encoder.transmission_p / control_feedback.primary_encoder.transmission_q;
-            ctrl.motor_current_constant = 1 / (parameter_scalar_get(&motor_params.torque_cst) / transmission);
+            ctrl.motor_current_constant = 1.f / (parameter_scalar_get(&motor_params.torque_cst) * transmission);
```
antoinealb: I think this is right.

FlorianKaufmann: Correct if transmission > 1

SyrianSpock:
Refactoring the transmission factors to use `uint32_t` seems tricky though, we rely on `uint16_t` overflow in the computations, needs more thorough investigation. I'm gonna merge this to advance on the arm control.

PR #203, opened 2019-04-19
Robot position live plot

Lu-ni:
- Send x,y,a over ethernet (via udp).
- Plot position of the robot with the python script 

You need to set you ethernet ip to 192.168.3.1. (example in fish):
```while true; sudo ifconfig enp0s25 192.168.3.1; sleep 1; end```
___
plot-xya

Improvement will be:
- to send udp packet via the raspi
- have both robot
- have puke

### CODE REVIEW

File: master-firmware/protobuf/position.proto [hunk 1]
```diff
@@ -0,0 +1,10 @@
+syntax = "proto2";
+
+import "nanopb.proto";
+
+message RobotPosition {
+    option (nanopb_msgopt).msgid = 10;
```
SyrianSpock: This will conflict with the arm state message https://github.com/cvra/robot-software/blob/master/master-firmware/protobuf/manipulator.proto#L6
13 should be fine

tools/log_udp_protobuf/plot_position.py
```diff
@@ -0,0 +1,123 @@
...
+                    'fill': 'grey'
+                },
+                'Experience1' : {
+                    'pts' : [(0,0),(0,-222),(450,-222),(450,0)],
+                    'fill': 'grey'
+                },
+                'Experience2' : {
+                    'pts' : [(2550,0),(2550,-222),(3000,-222),(3000,0)],
+                    'fill': 'grey'
+                },
```
SyrianSpock: The static elements (starting zones, ramp, central beacon holder) could be set once when we initialize `data` at line 33

tools/log_udp_protobuf/plot_position.py
```diff
@@ -0,0 +1,123 @@
...
+                'robot': {
+                    'x': msg.x,
+                    'y': msg.y,
+                    'a': (msg.a*180)/3.14,
```
SyrianSpock:
```py
from math import radians

'a': radians(msg.a)
```

PR #204, opened 2019-04-23
Read IP parameters from tree

antoinealb:
This PR implements two different things, both needed to have IP comms between robots:

* The MAC address is derived from hardware serial numbers and used a locally administered MAC.
* The IP is read from the config tree.

This means each robot will have a different IP address.

Tested that the IP address changes in lwip.

### DIFF
--- a/config_order.yaml
+++ b/config_order.yaml
@@ -4,6 +4,13 @@ logging:
     level: "notice"
   sdcard:
     level: "notice" # changing this to debug slows down things too much and causes panic
+ip:
+  ethernet:
+    address: "192.168.3.20"
+    netmask: "255.255.255.0"
+  uwb:
+    address: "10.0.0.20"
+    netmask: "255.255.255.0"
 master:
     is_main_robot: true
     control_panel_active_high: false
--- a/master-firmware/src/lwip_bindings/lwipthread.c
+++ b/master-firmware/src/lwip_bindings/lwipthread.c
@@ -58,6 +58,8 @@ limitations under the License.
 #include "hal.h"
 #include "evtimer.h"
 
+#include <crc/crc16.h>
+
 #include "lwipthread.h"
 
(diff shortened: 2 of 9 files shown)


config_order.yaml
```diff
@@ -4,6 +4,13 @@ logging:
     level: "notice"
   sdcard:
     level: "notice" # changing this to debug slows down things too much and causes panic
+ip:
+  ethernet:
+    address: "192.168.3.20"
```
SyrianSpock: [nitpicking] mixing double quotes and single quotes in the other robot's config file

master-firmware/src/lwip_bindings/lwipthread.c
```diff
@@ -69,19 +69,40 @@ limitations under the License.
 #include <lwip/stats.h>
 #include <lwip/snmp.h>
 #include <lwip/tcpip.h>
+#include <parameter/parameter.h>
 #include "netif/etharp.h"
 #include "netif/ppp/pppoe.h"
 #include "can/can_uwb_ip_netif.hpp"
+#include "main.h"
+#include "uid.h"
+#include "crc/crc16.h"
```
SyrianSpock: [nitpicking] `#include <crc/crc16.h>` and move above since it's from the `lib` folder

master-firmware/src/lwip_bindings/lwipthread.c
```diff
@@ -214,6 +235,13 @@ void ipinit_done_cb(void* a)
     sys_sem_signal(&lwip_init_done);
 }
 
+static uint16_t local_id(void)
+{
+    uint8_t id[UID_LENGTH];
+    uid_read(id);
+    return crc16(0xcafe, id, sizeof(id));
```
SyrianSpock: :coffee:

PR #206, opened 2019-04-25
goap : launch accelerator

Lu-ni:


master-firmware/src/strategy.cpp
```diff
@@ -319,6 +319,37 @@ struct DepositPuck : actions::DepositPuck {
...
+
+        if (!strategy_goto_avoid(MIRROR_X(m_color, 1305), 300, MIRROR_A(m_color, 90), TRAJ_FLAGS_ALL)) {
+            return false;
+        }
+        strategy_wait_ms(100);
+
+        trajectory_a_rel(&robot.traj, 13);
+        trajectory_wait_for_end(TRAJ_FLAGS_ROTATION);
+
+        state.arms_are_deployed = true;
```
antoinealb: Should be before `manipulator_goto`

PR #208, opened 2019-04-26
Collect the pucks in front of the start areas

SyrianSpock:
With this we should able to pick and deposit the pucks in front of the start areas (2 red and 1 green).

### DIFF
--- a/master-firmware/tests/test_strategy.cpp
+++ b/master-firmware/tests/test_strategy.cpp
@@ -1,30 +1,29 @@
 #include <CppUTest/TestHarness.h>
 #include <vector>
 #include <array>
-#include <iostream>
 
 #include "strategy/goals.h"
 #include "strategy/actions.h"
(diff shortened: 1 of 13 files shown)


master-firmware/tests/test_strategy.cpp
```diff
@@ -1,49 +1,57 @@
...
 #include <array>
+#include <iostream>
 
 #include "strategy/goals.h"
 #include "strategy/actions.h"
 
-bool dummy_execute(goap::Action<RobotState>* action, RobotState& state)
+bool dummy_execute(goap::Action<RobotState>* action, RobotState& state, const std::string& name)
 {
+    std::cout << name << std::endl;
```
antoinealb: intentional ?

SyrianSpock: nope, reverted

PR #209, opened 2019-04-26
Electron integration

antoinealb:
This exposes the following features:

* on boot, electron goes down until it meets its base
* when sent a message over UWB, it starts climbing until it reaches the top
* a shell command sends a command from master firmware


electron-firmware/README.md
```diff
@@ -0,0 +1,8 @@
+# CAN IO firmware
```
SyrianSpock: [typo] should be `Electron firmware`

electron-firmware/src/uavcan/node.cpp
```diff
@@ -0,0 +1,93 @@
...
+#define UAVCAN_CAN_BITRATE      1000000UL
+#define UAVCAN_MEMORY_POOL_SIZE 4096
+
+namespace uavcan_node {
+
+using Node = uavcan::Node<UAVCAN_MEMORY_POOL_SIZE>;
+using CanInterface = uavcan_stm32::CanInitHelper<UAVCAN_RX_QUEUE_SIZE>;
+using SystemClock = uavcan_stm32::SystemClock;
+
+/* TODO: Check which one is front */
```
SyrianSpock: still to be checked?

PR #212, opened 2019-04-27
Take even more pucks

SyrianSpock:
This PR opens the door for collecting even more pucks!
Changes include:
- Consider the pucks in front of the start position as obstacles until collection
- Avoid main static obstacles on the table (ramp and wall)
- Pick vertical pucks (tested with one blue puck)

### DIFF
--- a/master-firmware/src/base/map.c
+++ b/master-firmware/src/base/map.c
@@ -37,6 +37,10 @@ void map_init(struct _map* map, int robot_size)
     }
     map->last_opponent_index = 0;
 
+    /* Add the wall separating the two balances*/
+    map->the_wall = oa_new_poly(&map->oa, 4);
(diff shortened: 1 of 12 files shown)


master-firmware/src/base/map.c
```diff
@@ -37,29 +38,18 @@ void map_init(struct _map* map, int robot_size)
     }
     map->last_opponent_index = 0;
 
-    /* Add wastewater obstacle */
-    map->wastewater_obstacle = oa_new_poly(&map->oa, 4);
-    map_set_rectangular_obstacle(map->wastewater_obstacle, 1500, 1875, 1212, 250, robot_size);
+    /* Add the wall separating the two balances*/
+    map->ramp_obstacle = oa_new_poly(&map->oa, 4);
```
antoinealb: Shouldn't this be `map->the_wall` ?

SyrianSpock: oups!

PR #214, opened 2019-04-28
Proximity beacon speed control

antoinealb:
This PR implements proximity beacon speed control using the encoder built into the motor. It also adds filters to the quadrature encoder inputs, which prevents issues with the (i assume) noisy signal coming from the motors.

We should now test it on the robot, but no rush and please review it carefully, as I dont want it to cause encoder damage or loss of precision.

### DIFF
--- a/motor-control-firmware/src/encoder.c
+++ b/motor-control-firmware/src/encoder.c
@@ -22,6 +22,9 @@ void encoder_init_primary(void)
     STM32_TIM4->SMCR = STM32_TIM_SMCR_SMS(3); // count on both edges
     STM32_TIM4->CCMR1 = STM32_TIM_CCMR1_CC1S(1); // CC1 channel is input, IC1 is mapped on TI1
     STM32_TIM4->CCMR1 |= STM32_TIM_CCMR1_CC2S(1); // CC2 channel is input, IC2 is mapped on TI2
+    /* enable filtering, clock prescaler 32, 8 samples */
+    STM32_TIM4->CCMR1 |= STM32_TIM_CCMR1_IC1F(0xf);
+    STM32_TIM4->CCMR1 |= STM32_TIM_CCMR1_IC2F(0xf);
     STM32_TIM4->CCER = 0;
     STM32_TIM4->ARR = 0xFFFF;
     STM32_TIM4->CR1 = 1; // start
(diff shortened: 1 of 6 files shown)


proximity-beacon-firmware/src/proximity_beacon_publisher.cpp
```diff
@@ -19,8 +19,9 @@ static void timer_cb(const uavcan::TimerEvent& event)
         proximity_beacon_signal_delete(pbs);
     }
 
-    /* periodic setpoint update to keep the motor running */
-    control_update_voltage_setpoint(-4.0);
+    /* Rotate at 10 Hz */
+    const auto setpoint = -10 * 2 * 3.14f;
```
SyrianSpock:
```suggestion
    const float setpoint = -10 * 2 * M_PI;
```

antoinealb: M_PI is not a part of the C++ standard unfortunately...

FlorianKaufmann: It works if you add

#define _USE_MATH_DEFINES
#include <math.h>

SyrianSpock: It's sad that we have to use an internal define to use `M_PI` and there is no `math::pi<float>()` or equivalent C++-ish way of using pi. My sorrow knows no bounds 
sad_christmas

antoinealb:
Rebased & fixed build

motor-control-firmware/src/encoder.c
```diff
@@ -22,6 +22,9 @@ void encoder_init_primary(void)
     STM32_TIM4->SMCR = STM32_TIM_SMCR_SMS(3); // count on both edges
     STM32_TIM4->CCMR1 = STM32_TIM_CCMR1_CC1S(1); // CC1 channel is input, IC1 is mapped on TI1
     STM32_TIM4->CCMR1 |= STM32_TIM_CCMR1_CC2S(1); // CC2 channel is input, IC2 is mapped on TI2
+    /* enable filtering, clock prescaler 32, 8 samples */
+    STM32_TIM4->CCMR1 |= STM32_TIM_CCMR1_IC1F(0xf);
```
Lu-ni: What |= do ?

antoinealb: `a |= b` means `a = a | b` with `|` being the bitwise OR operator. Here it is used to set some bits:

Suppose you have a variable `foo` which holds the binary value `100` and you want to set the last bit to `1`.
You can do that by doing `foo |= 1;`, here is what the OR will look like:

```
  100
| 001
-----
  101 <- last bit is set
```

Here we use it to set some bits in CCMR1 to 0xf. To know the signification of those bits, you can look in the F3 reference manual, under the chapter for timers.

Lu-ni: Nice, thank you :)

antoinealb: Can you point to examples where there is a lot of preprocessor code ? Usually we use macros as a way to make code more readable. For example, the code above (`STM32_TIM_CCMR1_IC1F(0xf);`) is better than `0xf0` because it allows you to match the value with what you see in the datasheet.

Other than that we often have stuff like `#ifdef` to enable or disable features at compile time. The most common version of that is to build with or without the bootloader enabled, like [here](https://github.com/cvra/robot-software/blob/a5061d965ffeeca5f8466b65f7961a8b3951ac1f/motor-control-firmware/src/bootloader_config.c#L6-L31).

Finally, we sometimes have to write code using the preprocessor rather than simple C functions because there is simply no other way. One example of this is our [logging macros](https://github.com/cvra/robot-software/blob/a5061d965ffeeca5f8466b65f7961a8b3951ac1f/lib/error/error.h#L79-L88): The logging system needs to find the name of the source file and the current line in the file, in order to print those useful informations to the user. Unfortunately, such informations are not possible to find at any other time than preprocessing time: the line number will have changed once pre-preprocessing is finished for example.

Hope that explains a bit why we do it :)

PR #215, opened 2019-04-29
Classify more pucks

SyrianSpock:
I managed to make my code from Saturday night work :tada: however it's using more memory than before :crying_cat_face: . I'm not sure this is the right way to go with the goal/action definitions.

### DIFF
--- a/master-firmware/protobuf/strategy.proto
+++ b/master-firmware/protobuf/strategy.proto
@@ -16,14 +16,15 @@ enum PuckColor {
 
 message RobotState {
     option (nanopb_msgopt).msgid = 14;
+    option (nanopb_msgopt).packed_struct = true;
     required bool arms_are_indexed = 1 [default=false];
     required bool arms_are_deployed = 2 [default=true];
-    repeated bool puck_available = 3 [(nanopb).fixed_count=true, (nanopb).max_count=12, packed=true];
+    repeated bool puck_available = 3 [(nanopb).fixed_count=true, (nanopb).max_count=12];
     required bool has_puck = 4 [default=false];
-    required PuckColor has_puck_color = 5;
-    required bool accelerator_is_done = 6 [default=false];
-    repeated uint32 classified_pucks = 7 [(nanopb).fixed_count=true, (nanopb).max_count=4, packed=true];
-    required bool electron_launched = 8 [default=false];
-    required bool goldonium_in_house = 9 [default=true];
-    required bool has_goldonium = 10 [default=false];
+    required bool accelerator_is_done = 5 [default=false];
+    required bool electron_launched = 6 [default=false];
+    required bool goldonium_in_house = 7 [default=true];
+    required bool has_goldonium = 8 [default=false];
+    required PuckColor has_puck_color = 9;
+    repeated uint32 classified_pucks = 10 [(nanopb).fixed_count=true, (nanopb).max_count=4, packed=true];
 }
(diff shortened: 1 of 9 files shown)


master-firmware/protobuf/strategy.proto
```diff
@@ -11,18 +11,18 @@ enum PuckColor {
     RED = 0;
     GREEN = 1;
     BLUE = 2;
-    WHITE = 3;
+    REDGREEN = 3;
```
antoinealb: nitpick: add comment specifying that it means red or green, aka unknown ?

master-firmware/src/strategy/actions.h
```diff
@@ -6,7 +6,7 @@
 
 /** Number of states goap can visit before giving up. Increasing it means a
  * solution is found on more complex problems at the expense of RAM use. */
-#define GOAP_SPACE_SIZE 40
+#define GOAP_SPACE_SIZE 180
```
antoinealb: omfg, so expensive.

master-firmware/src/strategy/state.cpp
```diff
@@ -6,10 +6,13 @@ RobotState initial_state(void)
 {
     RobotState state;
     state = RobotState_init_default;
-    for (size_t i = 0; i < 3; i++)
+
+    size_t num_pucks = sizeof(state.puck_available) / sizeof(bool);
```
antoinealb: Does this work in a packed context ? Not sure.

master-firmware/protobuf/strategy.proto
```diff
@@ -11,18 +11,18 @@ enum PuckColor {
...
 }
 
 message RobotState {
     option (nanopb_msgopt).msgid = 14;
```
antoinealb: I would suggest adding `option (nanopb_msgopt).packed_struct = true;` to RobotState. Using that and reordering the fields to have all the `bool` together (for optimal packing), I got the state size from 32 to 26 bytes. Might not seem much, but every edge matters.

master-firmware/src/strategy/goals.h
```diff
@@ -11,24 +11,31 @@ struct InitGoal : goap::Goal<RobotState> {
     }
 };
 
-struct RedPucksGoal : goap::Goal<RobotState> {
+struct ClassifyStartPucksGoal : goap::Goal<RobotState> {
     virtual int distance_to(const RobotState& state) const
     {
-        return goap::Distance().shouldBeEqual(state.pucks_in_deposit_zone[PuckColor_RED], 2);
+        return goap::Distance().shouldBeEqual(state.classified_pucks[PuckColor_REDGREEN], 3);
```
antoinealb: Caution, those types of goals will fail completely if the robot can only find 2 pucks for example.

master-firmware/protobuf/strategy.proto
SyrianSpock: what if I rename it `RED_OR_GREEN`?

master-firmware/src/strategy/goals.h
SyrianSpock: I agree, I'm going to change the goals again over this week

master-firmware/src/strategy/state.cpp
SyrianSpock: Looking at the generated code, it's a regular C array of booleans
```
bool puck_available[12];
```
so it's fine. Also the `packed=true` doesn't change the generated code, so I guess it's redundant.

PR #216, opened 2019-04-30
Electron gets smarter

nuft:
Some little improvements of the electron:
- add timeout for reaching target
- comunicate state through LED blink pattern

This change is tested and working on hardware.

### DIFF
--- a/electron-firmware/src/uavcan/node.cpp
+++ b/electron-firmware/src/uavcan/node.cpp
@@ -19,6 +19,19 @@ using Node = uavcan::Node<UAVCAN_MEMORY_POOL_SIZE>;
 using CanInterface = uavcan_stm32::CanInitHelper<UAVCAN_RX_QUEUE_SIZE>;
 using SystemClock = uavcan_stm32::SystemClock;
 
+static void timeout_cb(void *flag_p)
+{
+    *(bool*)flag_p = true;
+}
+
+static void start_timer(unsigned int timeout_sec, bool* flag_p)
+{
+    static virtual_timer_t timer;
+    const sysinterval_t timeout = OSAL_S2I(timeout_sec);
+    chVTObjectInit(&timer);
+    chVTDoSetI(&timer, timeout, timeout_cb, flag_p);
+}
+
 void main(unsigned int id, const char* name)
 {
     int res;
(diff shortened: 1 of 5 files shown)


electron-firmware/src/uavcan/node.cpp
```diff
@@ -60,14 +73,35 @@ void main(unsigned int id, const char* name)
...
+        }
+
+        if (electron_state == RUNNING) {
+            if (data_packet_start_signal_received() && !front_hall_sensor()) {
+                motor_voltage_set(voltage);
+            } else {
+                motor_voltage_set(0.0f);
+            }
+
+            if (front_hall_sensor() && timed_out) {
```
antoinealb: Shouldn't it be `front_hall || timed_out` ?

electron-firmware/src/uavcan/node.h
```diff
@@ -5,6 +5,14 @@
 extern "C" {
 #endif
 
+enum {
+    INIT,
+    READY,
+    RUNNING,
+    ARRIVED
+};
```
SyrianSpock: Could be a `enum class ElectronState {...}` for type safety, then you'll need to use `ElectronState::INIT`

electron-firmware/src/uavcan/node.cpp
SyrianSpock: I thought the idea was to continue climbing until timing out, but I'm not sure anymore what we're trying to achieve :thinking:

nuft: The idea was that it continues to try to climb until the timeout.
I realize now that it should only be `if (timed_out)`.

electron-firmware/src/uavcan/node.h
nuft: I want it to be usable from `main.c` and I'm too lazy to make it a `main.cpp`.

electron-firmware/src/uavcan/node.cpp
nuft: changed now. I added a TIMED_OUT state that should allow to define success only when in ARRIVED state. We could add a message back to the robot for exact score counting.

electron-firmware/src/uavcan/node.h
SyrianSpock: fair enough

PR #219, opened 2019-05-05
Ctrl

FlorianKaufmann:
This pull request implements a LQR controller to regulate the arms.

The LQR controller is deployed in the manupulator_thread.cpp. It is possible to switch from PID to LQR by changing the ctrl flag which is for now defined in the config file.

The LQR controller takes as input the 3 target angles. The references to the target r(t) and v(t) are computed in RefTracking.cpp. The output of the LQR controller is an angular acceleration vector. It is transformed into torque and compensation for gravity is added to optimize the control in function of the arm position.


antoinealb:
Closing because obsolete.

PR #222, opened 2019-05-07
Add a service that sends position to ally over UDP

antoinealb:
Since we have UDP over UWB (UoU) we now effectively have our position
sent over UWB. I need to write a client (receive side) for it.

How to test: Flash it, and then change the `/ip/ally_address config` key to point to your laptop (either over UWB/CAN or via Ethernet). You can then use `tools/log_udp_protobuf/log_ally_position.py` script to listen for position messages.

@SyrianSpock You can either merge as-is or once I implement the RX side. I have no time to do it today, but it should be done tomorrow night.


antoinealb:
So this is now complete, with both sending the position to the other robot and receiving position information and publishing it on `/ally_pos`. Was tested on my desk, using Ethernet instead of UWB, and it worked. So if @SyrianSpock or anybody wants to integration test it, please do, I dont have access to the bot.

PR #223, opened 2019-05-07
export graph diagram from manipulator.h

FlorianKaufmann:
This PR implements a python script to visualize the connectivity graph of manipulator.h and eventually other files with similar syntax.


Lu-ni:
Juste move it to /robot-software/tools/
Otherway, LGTM !

PR #232, opened 2019-05-31
[WIP] Get out of obstacle

SyrianSpock:
Still a work in progress, but this PR does the following:
- switches motion planning to happen asynchronously in a thread
- compute a plan all the time
- if stuck in an opponent, move radially away from it

First tests seem OK, but it requires more thorough testing, especially on the strategy, that going somewhere actually goes where we specify. In the meantime, a first look can't hurt ;)


master-firmware/src/strategy.cpp
```diff
@@ -68,13 +68,43 @@ static void strategy_forward(void* ctx, int relative_distance_mm)
...
     strategy_context_t* strat = (strategy_context_t*)ctx;
-    return strategy_goto_avoid(strat, x_mm, y_mm, a_deg, TRAJ_FLAGS_ALL);
+
+    motion_planner_set_goal(x_mm, y_mm, a_deg);
+    auto end_reason_xy = trajectory_wait_for_end(TRAJ_FLAGS_ALL);
+
+    if (end_reason_xy == TRAJ_END_GOAL_REACHED) {
+        // wait again for rotation
+        strat->wait_ms(200);
+        auto end_reason_a = trajectory_wait_for_end(TRAJ_FLAGS_ALL);
```
nuft: Why do you wait again when the end_reason doesn't change?

master-firmware/src/strategy.cpp
```diff
@@ -362,11 +392,12 @@ void strategy_play_game(void* p)
     NOTICE("Waiting for color selection...");
     strategy.color = wait_for_color_selection();
     map_server_start(strategy.color);
+    motion_planner_start();
     score_counter_start();
 
     if (config_get_boolean("master/is_main_robot")) {
         NOTICE("First, Order...");
-        strategy_order_play_game(&strategy, state);
+        strategy_chaos_play_game(&strategy, state);
```
nuft: don't forget to change back

master-firmware/src/robot_helpers/trajectory_helpers.c
```diff
@@ -63,7 +63,7 @@ int trajectory_has_ended(int watched_end_reasons)
         if (messagebus_topic_read(proximity_beacon_topic, &beacon_signal, sizeof(beacon_signal)) && timestamp_duration_s(beacon_signal.timestamp.us, timestamp_get()) < TRAJ_MAX_TIME_DELAY_OPPONENT_DETECTION && beacon_signal.range.range.distance < TRAJ_MIN_DISTANCE_TO_OPPONENT) {
             float x_opp, y_opp;
             beacon_cartesian_convert(&robot.pos,
-                                     1000 * beacon_signal.range.range.distance,
+                                     1.25 * 1000 * beacon_signal.range.range.distance,
```
nuft: This correction should be done somewhere else.
Also, the error is an offset not a factor.

master-firmware/src/base/base_controller.c
```diff
@@ -264,3 +267,73 @@ void trajectory_manager_start(void)
...
     chThdCreateStatic(trajectory_thd_wa, sizeof(trajectory_thd_wa), TRAJECTORY_MANAGER_PRIO, trajectory_manager_thd, &(robot.traj));
 }
+
+static struct {
+    int x_mm;
+    int y_mm;
+    int a_deg;
+} goal;
+
+void motion_planner_set_goal(int x_mm, int y_mm, int a_deg)
```
nuft: no locking?

master-firmware/src/strategy.cpp
SyrianSpock: it's mentioned in the comments, we need to wait for the final rotation, still don't like this, it's flaky IMO

master-firmware/src/strategy.cpp
SyrianSpock: oups shouldn't have committed

master-firmware/src/robot_helpers/trajectory_helpers.c
SyrianSpock: euh, this is not what I wanted to do :man_facepalming: 
I need to scale the opponent obstacle size here

master-firmware/src/base/base_controller.c
SyrianSpock: :bomb:

master-firmware/src/strategy.cpp
nuft: Do I understand right that `end_reason_a` is `TRAJ_END_GOAL_REACHED` independent of the angle? (depends only on xy)

SyrianSpock: `TRAJ_END_GOAL_REACHED` is returned when either of trajectory setpoints is reached (on distance and angle) this covers the last xy setpoint before reaching a goal and the last rotation we do after reaching the xy setpoint.
see https://github.com/cvra/robot-software/blob/master/master-firmware/src/robot_helpers/trajectory_helpers.c#L42-L44

SyrianSpock:
Note: this will make debugging wheelbase motion harder but it opens the way for more flexible strategy actions

master-firmware/src/base/base_controller.c
```diff
@@ -310,6 +311,10 @@ static THD_FUNCTION(motion_planner_thd, arg)
             trajectory_goto_xy_abs(&robot.traj, points[0].x, points[0].y);
         } else {
             WARNING("No path found!");
+            poly_t* opponent = map_get_opponent_obstacle(map, 0);
```
antoinealb: This only checks for a single opponent, correct ?

master-firmware/src/robot_helpers/trajectory_helpers.c
```diff
@@ -125,6 +125,11 @@ static bool trajectory_is_cartesian(struct trajectory* traj)
     return traj->state == RUNNING_XY_START || traj->state == RUNNING_XY_ANGLE || traj->state == RUNNING_XY_ANGLE_OK || traj->state == RUNNING_XY_F_START || traj->state == RUNNING_XY_F_ANGLE || traj->state == RUNNING_XY_F_ANGLE_OK || traj->state == RUNNING_XY_B_START || traj->state == RUNNING_XY_B_ANGLE || traj->state == RUNNING_XY_B_ANGLE_OK;
 }
 
+bool trajectory_robot_inside_obstacle(struct _robot* robot, poly_t* opponent)
+{
+    return math_point_is_in_square(opponent, position_get_x_s16(&robot->pos), position_get_y_s16(&robot->pos));
```
antoinealb: Name is misleading, is_in_square should be is_in_poly.

master-firmware/src/base/base_controller.c
```diff
@@ -264,3 +266,58 @@ void trajectory_manager_start(void)
...
+{
+    (void)arg;
+    chRegSetThreadName(__FUNCTION__);
+
+    while (1) {
+        chThdSleepMilliseconds(100);
+
+        const point_t start = {position_get_x_float(&robot.pos), position_get_y_float(&robot.pos)};
+        if (((start.x - goal.x_mm) * (start.x - goal.x_mm) + (start.y - goal.y_mm) * (start.y - goal.y_mm)) < 100) {
+            NOTICE("Close enough!");
```
antoinealb: This will spam the logs with NOTICE at 10 Hz, thats not really great from a debug point of view.

master-firmware/src/base/base_controller.c
```diff
@@ -264,3 +266,58 @@ void trajectory_manager_start(void)
...
+        oa_start_end_points(&map->oa, start.x, start.y, goal.x_mm, goal.y_mm);
+        oa_process(&map->oa);
+
+        /* Retrieve path */
+        point_t* points;
+        int num_points = oa_get_path(&map->oa, &points);
+        DEBUG("Path to (%d, %d) computed with %d points", goal.x_mm, goal.y_mm, num_points);
+
+        if (num_points == 0) {
+            NOTICE("Done!");
```
antoinealb: num_points will most likely never be the case. You will never be perfectly at the end of the path, so I think path should always be at least of length 1.

master-firmware/src/base/base_controller.c
```diff
@@ -264,3 +266,58 @@ void trajectory_manager_start(void)
...
+        } else if (num_points > 0) {
+            trajectory_goto_xy_abs(&robot.traj, points[0].x, points[0].y);
```
antoinealb: This is pretty problematic. The current design means it is not possible anymore to use the trajectory module in the "go forward x mm" mode or "turn 90 deg" mode anymore, as this will always be overwritten by the motion manager thread.

master-firmware/src/robot_helpers/trajectory_helpers.c
SyrianSpock: true but here we assume for the sake of simplicity the opponent is a square

master-firmware/src/base/base_controller.c
SyrianSpock: :+1:

SyrianSpock:
I think we can safely decline this PR it's a suboptimal incomplete solution to the problem, but we get a 1st idea of how not to tackle it :wink:

PR #233, opened 2019-06-09
[RFC, WIP] Cmake build infra

antoinealb:
This is the current status of my work on the CMake integration.  It is not really ready to be used, but I wanted to get feedback as early as possible.
For now only the UWB beacon is working, and I am expecting to fix the others pretty soon.

I encourage everyone to try it, and see if you like the CMake workflow.

@SyrianSpock, I think you are the most experienced with CMake, so I would be happy to have a review from you on the cmake infra itself.

Summary of the changes:

1. The includes for modules now live in a `include` directory for each library.
    For example the includes for the PID filters live under `lib/pid/include/pid`.
    This avoids having accidental dependencies: it is not impossible to include a header from a non-linked library.
2. Each module has its own test executable, and all the executables are built together.
    Before each platform had its own set of tests, and they all went in the same exec.
    It is now easier to have other tests, such as integration tests or tests from third party libraries running.

Things missing (probably not exhaustive):

* CAN bootloader integration
* UAVCAN messages are sometimes recompiled, sometimes not.
    Still working on it.
* Only UWB firmware builds
* For now it is not possible to build a board with and without the bootloader.
    Is it something that we really need ?
    In practice I feel that we are always using the same config (without bootloader for master, with bootloader for motor). 
* Continuous integration, which is now botched up..
* Testing on robot
    

Advantages of using cmake:

* It is possible to use `CMAKE_EXPORT_COMPILE_COMMANDS` to generate Clang's compile command database.
    This in turn can be used for smart autocompletes, linters, etc.
* Faster build, especially incremental build (havent measured the progress, but feels faster).
* We now own 100% of the build system and do not rely on ChibiOS for it.
* Ability to generate IDE projects (Eclipse, Xcode) for those who would like it.
* No more `packager` that can be forgotten to run.


antoinealb:
Friendly ping @SyrianSpock 👼

.travis.yml
```diff
@@ -1,21 +1,17 @@
...
         sources:
-            - deadsnakes
             - ubuntu-toolchain-r-test
         packages:
             - python3.5
             - python3.5-dev
             - python-pip
             - python-virtualenv
-            - gcc-4.9
-            - g++-4.9
```
SyrianSpock: scary compiler upgrade, 
3573mf

README.md
```diff
@@ -29,6 +29,27 @@ Other important software components can be found in this repo:
...
+# The resulting firmware is at `uwb-beacon-firmware/uwb-beacon-firmware.elf`.
+```
+
+To build the unit tests for all the boards:
+
+```bash
+mkdir build
+cd build
+cmake ..
+make all test
```
SyrianSpock: so we have to call `all` all the time?

cmake/gcc_stm32f0.cmake
```diff
@@ -0,0 +1,94 @@
...
+
+SET(CMAKE_EXE_LINKER_FLAGS "-Wl,--gc-sections -mthumb -mcpu=cortex-m0 -mabi=aapcs" CACHE INTERNAL "executable linker flags")
+SET(CMAKE_MODULE_LINKER_FLAGS "-mthumb -mcpu=cortex-m0 -mabi=aapcs" CACHE INTERNAL "module linker flags")
+SET(CMAKE_SHARED_LINKER_FLAGS "-mthumb -mcpu=cortex-m0 -mabi=aapcs" CACHE INTERNAL "shared linker flags")
+
+SET(STM32_CHIP_TYPES 030x6 030x8 031x6 038xx 042x6 048x6 051x8 058xx 070x6 070xB 071xB 072xB 078xx 091xC 098xx 030xC CACHE INTERNAL "stm32f0 chip types")
+SET(STM32_CODES "030.[46]" "030.8" "031.[46]" "038.6" "042.[46]" "048.6" "051.[468]" "058.8" "070.6" "070.B" "071.[8B]" "072.[8B]" "078.B" "091.[BC]" "098.C" "030.C")
+
+MACRO(STM32_GET_CHIP_TYPE CHIP CHIP_TYPE)
+    STRING(REGEX REPLACE "^[sS][tT][mM]32[fF]((03[018].[468C])|(04[28].[46])|(05[18].[468])|(07[0128].[68B])|(09[18].[BC])).*$" "\\1" STM32_CODE ${CHIP})
```
SyrianSpock: That's a monster right there, I heard they use it to scare children :scream_cat:

cmake/gcc_stm32f0.cmake
```diff
@@ -0,0 +1,94 @@
+SET(CMAKE_C_FLAGS "-mthumb -fno-builtin -mcpu=cortex-m0 -Wall -std=gnu99 -ffunction-sections -fdata-sections -fomit-frame-pointer -mabi=aapcs -fno-unroll-loops -ffast-math -ftree-vectorize" CACHE INTERNAL "c compiler flags")
```
SyrianSpock: SO WE'RE GOING FOR ALL CAPS? :sweat_smile:

lib/cmp/test/test.c
```diff
@@ -9,7 +9,7 @@
 #include "utils.h"
 #include "buf.h"
 
-#include "cmp.h"
+#include "cmp/cmp.h"
```
SyrianSpock: <> get no love?

lib/cmp/test/test.c
```diff
@@ -9,7 +9,7 @@
 #include "utils.h"
 #include "buf.h"
```
SyrianSpock: this memory is so buf

lib/crc/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+add_library(crc
+crc32.c
+crc16.c
+)
```
SyrianSpock: no indent for filenames?

lib/crc/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+target_include_directories(crc PUBLIC include)
+
+
+
+cvra_add_test(TARGET crc_test SOURCES 
+    test/crc32_test.cpp
+    test/crc16_test.cpp
+    DEPENDENCIES
+    crc
```
SyrianSpock: isn't that beautiful?

lib/filter/iir.c
```diff
@@ -1,7 +1,7 @@
 #include <math.h>
 #include <string.h>
 
-#include "iir.h"
+#include "filter/iir.h"
```
SyrianSpock: I'd love to see more <> on those public header includes

lib/filter/tests/basic_test.cpp
```diff
@@ -1,5 +1,5 @@
 #include "CppUTest/TestHarness.h"
-#include "../basic.h"
```
SyrianSpock: goodbye `..` you will not be missed

lib/parameter/package.yml
```diff
@@ -4,10 +4,10 @@ source:
     - parameter_print.c
 
 depends:
-    - cmp_mem_access:
-        fork: stapelzeiger
```
SyrianSpock: bye bye the fork, you've served us well

lib/parameter/package.yml
```diff
@@ -4,10 +4,10 @@ source:
...
+    - cmp_mem_access
     - test-runner
     - cmp
+    - cmp_mem_access
```
SyrianSpock: twice?

.travis.yml
antoinealb: Its only for unit test though. And they pass.

README.md
antoinealb: yup, because CTest does not know what to recompile apparently.. But you can also just build the tests for the module you are working on ("make pid_test test")

cmake/gcc_stm32f0.cmake
antoinealb: comes from https://github.com/ObKo/stm32-cmake/blob/master/LICENSE, should add this on top of the cmake

lib/cmp/test/test.c
antoinealb: will format this.

lib/parameter/package.yml
antoinealb: package.yml will be removed anyway, I may have comitted useless stuff

lib/filter/iir.c
antoinealb: Fixed in (b9f44270443ebf21de4eb6dd9f8a3c6f864f7794 might be rebased at any point). I also added some automation to enforce that those headers stay coherent.

lib/crc/CMakeLists.txt
antoinealb: it is. I love it.

CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+add_c_cxx_compiler_flag("-Wall")
+add_c_cxx_compiler_flag("-Wextra")
+add_c_cxx_compiler_flag("-Wundef")
+add_c_cxx_compiler_flag("-Wduplicated-branches")
+add_c_cxx_compiler_flag("-Wduplicated-cond")
+add_c_cxx_compiler_flag("-Wlogical-op")
+add_c_cxx_compiler_flag("-Wnull-dereference")
+add_c_cxx_compiler_flag("-Wshadow")
+add_c_compiler_flag("-Wstrict-prototypes")
+add_c_compiler_flag("-Wfoobar-prototypes")
```
SyrianSpock: Could be shortened to 
```
set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} -Wall....)
```

cmake/ChibiOS/18.2/ChibiOS_RTOS.cmake
```diff
@@ -0,0 +1,93 @@
...
+    os/common/portability/GCC
+    os/common/startup/ARMCMx/compilers/GCC
+    os/common/ext/ARM/CMSIS/Core/Include
+)
+
+SET (CHIBIOS_SOURCES_${CHIBIOS_KERNEL}_MAILBOXES  os/common/oslib/src/chmboxes.c)
+SET (CHIBIOS_SOURCES_${CHIBIOS_KERNEL}_MEMCORE    os/common/oslib/src/chmemcore.c)
+SET (CHIBIOS_SOURCES_${CHIBIOS_KERNEL}_HEAP       os/common/oslib/src/chheap.c)
+SET (CHIBIOS_SOURCES_${CHIBIOS_KERNEL}_MEMPOOLS   os/common/oslib/src/chmempools.c)
+SET (CHIBIOS_SOURCES_${CHIBIOS_KERNEL}_FACTORY    os/common/oslib/src/chfactory.c)
```
SyrianSpock: Not the easiest name 😅

lib/cmp_mem_access/example.c
```diff
@@ -3,7 +3,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <cmp/cmp.h>
-#include "cmp_mem_access.h"
```
SyrianSpock: Good

lib/lwip/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
+if (${CMAKE_CROSSCOMPILING})
+    add_library(lwip
+        lwip/src/api/api_lib.c
```
SyrianSpock: Weird indentation

CMakeLists.txt
antoinealb: Not exactly. Look in cmake/AddCCompilerFlag.cmake. Basically, the macro adds it to both C and CXX flags, and checks if the compiler allows it before adding it to the options. This allows us to support advanced checks on modern compilers and still build on old versions.

cmake/ChibiOS/18.2/ChibiOS_RTOS.cmake
antoinealb: Well it comes from the template I used :(

cmake/gcc_stm32f0.cmake
SyrianSpock: should we mention the link?

SyrianSpock:
Do we plan to merge soon?

nuft:
Nice, It builds! Does it run on hardware?

uwb-beacon-firmware/CMakeLists.txt
```diff
@@ -0,0 +1,158 @@
...
+    tests/lru_cache.cpp
+    DEPENDENCIES
+    lru_cache
+)
+
+add_library(mpu9250
+    src/mpu9250.c
+)
+
+target_include_directories(mpu9250 PUBLIC src)
```
nuft: Is it worth adding separate library targets for sources that are not (yet) used anywhere else?

uwb-beacon-firmware/STM32F405xG.ld
```diff
@@ -88,4 +88,7 @@ REGION_ALIAS("HEAP_RAM", ram0);
 _config_start = ORIGIN(config);
 _config_end = ORIGIN(config) + LENGTH(config);
 
+__process_stack_size__ = 0x400;
+__main_stack_size__ = 0x400;
```
nuft: This will be replaced by a linker flag? Isn't this already done in `ChibiOS_LD.cmake`?

antoinealb:
> Nice, It builds! Does it run on hardware?

It does ! At least tested for the uwb-beacon. Bootloader integration is not yet ported over though.

uwb-beacon-firmware/CMakeLists.txt
antoinealb: I am not sure what you mean. That target is used in a test and in a binary.

uwb-beacon-firmware/STM32F405xG.ld
antoinealb: For now its not a linker flag, I am not sure we are using `ChibiOS_LD.cmake`. I actually prefer having it in the linker script rather than in a flag, it feels cleaner and easier to review.

nuft: Ok, agree. It's better to define such parameters in the source than in build system files.

uwb-beacon-firmware/CMakeLists.txt
nuft: Ok, I missed that there was a unit test for it.

SyrianSpock:
So, what's the plan? Are we merging this?

antoinealb:
Well CI is failing, and only one board is ported to the new build system, so I guess the answer is "not yet".

Current status:

* I am working on porting master-firmware
* I believe @nuft is working on can-io-firmware
* CAN bootloader integration is not complete yet, but should not be too hard to do.

SyrianSpock:
Ok, so the project is still alive :+1:

antoinealb:
Yes, it got rebased this weekend, and many commits are still in my home laptop.

nuft:
I started with can-io in a branch (sould probably wok on this branch though). 
There I added bootloader config support (without touching the bootloader repo).
I’m struggling with version.c generation and linker issues with some missing syscalls.

antoinealb:
I can probably write version.c generation. I also had some issues with linker and syscalls but i don't remember how I fixed those. Fortunately, I will have to do it again for master-firmware.

antoinealb:
@nuft I included version.c generation on this branch. Please push to another branch as I will often force push here.

antoinealb:
@nuft you might want to link lib/chibios-syscall/newlib_syscalls.c in your build, helped a lot for me.

nuft:
@antoinealb is it possible that you forgot to commit cmake/master.cmake?
Or where do you define MCU, board name, compiler flags, etc for master-firmware?

antoinealb:
> @antoinealb is it possible that you forgot to commit cmake/master.cmake?
> Or where do you define MCU, board name, compiler flags, etc for master-firmware?

Yes its likely. You can use uwb-beacon.cmake as reference.

nuft:
> you might want to link lib/chibios-syscall/newlib_syscalls.c in your build, helped a lot for me.

yes, it helped. CAN-IO builds now but flash section is overflown. Need to investigate why the binary size blew up.

cmake/uwb-beacon.cmake
```diff
@@ -0,0 +1,75 @@
...
+set(CMAKE_AR "arm-none-eabi-ar" CACHE INTERNAL "ar")
+set(CMAKE_RANLIB "arm-none-eabi-ranlib" CACHE INTERNAL "ranlib")
+
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+
+# Set flags common to C and C++ such as hardware platform, ABI and so on
+set(HARDWARE_FLAGS "-mthumb -mcpu=cortex-m4 \
+    -mfpu=fpv4-sp-d16 -mfloat-abi=softfp -mabi=aapcs"
```
nuft: This should be `-mfloat-abi=hard` to avoid copying back and forth between fp and working registers.

nuft:
I managed to reduce the size, however I'm still 20k behind the previous build size.

I also observed that the chibios-syscalls library does not overwrite the default implementation (like `operator delete()` for example, which ends up calling the newlib implementation.
This could have something to do with the library order during linking but I'm not sure.
My temporary workaround was to build the sources directly without using the library archive.

cmake/uwb-beacon.cmake
antoinealb: Good catch, will change it.

antoinealb:
Can you try playing with visibility settings ? https://stackoverflow.com/questions/17080869/what-is-the-cmake-equivalent-to-gcc-fvisibility-hidden-when-controlling-the-e

I also built the sources directly for chibios-syscalls, which sucks. Maybe we can use an INTERFACE library for that, will need to dig down in it.

SyrianSpock:
Are we close to a mergeable state? I worry that soon, the code will change a lot and make you redo this PR work

nuft:
Already part of #273

PR #236, opened 2019-07-18
Switch master-firmware to lib/trace

### LINKED ISSUE #235: Migrate master-firmware tracing to lib/trace
So far we have two different tracing systems, which are almost identical. From a quick diff, it appears that the most important difference is that lib/trace is doing platform-specific stuff using functions where master firmware is doing that using preprocessor macro.

As preprocessor macros are harder to integrate in a clean build, I suggest removing the tracing code from master-firmware and using lib/trace.

antoinealb:
Fixes #235


antoinealb:
@Lu-ni please review

PR #237, opened 2019-07-18
ci: Do not distinguish between order & chaos

### LINKED ISSUE #234: Travis build still distinguishes between Order and Chaos

antoinealb:
Nowadays, both robots have the same firmware and only the loaded
configuration is changed at runtime depending on the presence of a
jumper.

Fixes #234 (Travis build still distinguishes between Order and Chaos)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,7 @@ matrix:
         - compiler: gcc
           env: BUILD_TYPE="tests"
         - compiler: gcc
-          env: BUILD_TYPE="build" PLATFORM="master-firmware" ROBOT="order"
-        - compiler: gcc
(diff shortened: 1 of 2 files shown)


antoinealb:
@Lu-ni please review

Lu-ni:
In the past, we were building two different .elf ? I can't find anything on that. But the line you removed suggested it
`make $ROBOT_FLAG`


Otherewise, LGTM !

antoinealb:
Yes, we used to flash two different elf, each only containing strat & config for one robot. It was super error prone so now this is handled at runtime using a microcontroller pin wired to zero on one robot and to 1 on the other.

Issue #240, opened 2019-10-01
Setup Motor Board Emulator project


antoinealb:
* UAVCAN
* CMake
* SocketCAN
* C++ based
* (gGRPC)

PR #246, opened 2019-10-04
Setup motor board emulator UAVCAN + Cmake

antoinealb:
This pull request set ups the base project for our simulation idea. It spins up a single node, giving it a CAN id and a CAN name based on the commandline flags.

It adds a dependency on Abseil, which I like as a stdlib extension after using it at work, but maybe we should discuss whether or not we are ok with this dependency.

We will most likely review this during the next software call.

Fixes issue #240 

Tested manually by running it on a virtual SocketCAN adapter and querying the bus with UAVCAN's monitoring tools.

### DIFF
--- a/hitl/logging.cpp
+++ b/hitl/logging.cpp
@@ -2,16 +2,24 @@
 #include <cstdarg>
 #include "error/error.h"
 #include "absl/flags/flag.h"
+#include "absl/synchronization/mutex.h"
 
 ABSL_FLAG(bool, verbose, false, "Enable verbose logging.");
+static ABSL_CONST_INIT absl::Mutex logging_lock(absl::kConstInit);
 
 static void log_message(struct error* e, ...)
 {
+    absl::MutexLock l(&logging_lock);
     va_list va;
     va_start(va, e);
     printf("%s ", error_severity_get_name(e->severity));
     vprintf(e->text, va);
+    printf("\n");
     va_end(va);
+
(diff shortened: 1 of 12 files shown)


antoinealb:
Tested today, and it is possible to have several UAVCAN instances in the same process (several threads). Therefore the whole gRPC design might not be needed.

@nuft Back to the whiteboard !

hitl/logging.cpp
```diff
@@ -0,0 +1,34 @@
+#include <cstdio>
+#include <cstdarg>
+#include "error/error.h"
+#include "absl/flags/flag.h"
+#include "absl/synchronization/mutex.h"
+
+ABSL_FLAG(bool, verbose, false, "Enable verbose logging.");
+static ABSL_CONST_INIT absl::Mutex logging_lock(absl::kConstInit);
```
nuft: why not use `std::mutex`?

nuft:
> Therefore the whole gRPC design might not be needed.

Interesting, we should keep it simple - at least for the first version.
So maybe keep everything in a single process?

hitl/logging.cpp
antoinealb: https://abseil.io/about/design/mutex

That being said it's mostly habit and we could use std mutex if you prefer.

antoinealb:
Yes I would advocate for a two process design: still separating the visualization.

hitl/logging.cpp
nuft: No, let's keep abseil

PR #247, opened 2019-10-17
Enable automatic clang formatting workflow

antoinealb:
This pull request enables automatic clang formatting of our source code. this guarantees an uniform code formatting on our whole codebase, without anyone having to remember how its done.

The code formatter will pull any new branch, run clang-format on it and push back if there are any changes. Look at this branch for example.


nuft:
Nice
I think it's good to enforce code formatting.
Another option would be to enforce it through the CI.

Issue #242, opened 2019-10-01
message bus topic injection via UDP


antoinealb:
See https://git.antoinealb.net/antoinealb/hydrocontest-fw/blob/master/src/telemetry/udp_topic_injector.c

antoinealb:
This will most likely not be needed, as we now have a simulator that is interface 100% over UAVCAN.

PR #248, opened 2019-10-17
[WIP] UDP topic injector

nuft:
Addresses #242
- [x] Add topic injector by @antoinealb 
- [x] Test topic broadcaster
- [x] Enable debug output UART3 over Nucleo144 ST-Link
- [ ] Test topic injector


master-firmware/src/udp_topic_injector.c
```diff
@@ -10,6 +10,9 @@ static void udp_topic_injector_thd(void *p)
 
     struct netconn *conn;
 
+    chThdSetPriority(HIGHPRIO);
```
antoinealb: Can you add a comment explaining why HIGHPRIO is required there ?

master-firmware/src/debug/log.c
```diff
@@ -13,7 +13,7 @@
 #include "usbconf.h"
 #include "gui.h"
 
-#define OUTPUT_STREAM ((BaseSequentialStream*)&SD7)
+#define OUTPUT_STREAM ((BaseSequentialStream*)&SD3)
```
antoinealb: What does SD7 & SD3 point to ?

master-firmware/src/main.c
```diff
@@ -185,6 +186,8 @@ int main(void)
 {
     /* Initializes a serial driver.  */
     sdStart(&SD7, &debug_uart_config);
+    palSetPadMode(GPIOD, GPIOD_PIN8, PAL_MODE_ALTERNATE(7) | PAL_STM32_PUPDR_PULLUP); // ST-Link UART3 TX
```
antoinealb: palSetMode should be replaced by changes in board.h in my opinion.

master-firmware/src/debug/log.c
antoinealb: Is this to debug using just a Nucleo, outside a robot?

antoinealb:
Rebased on current master for testing with #261 next time we get some time with hardware.

master-firmware/src/udp_topic_injector.c
SyrianSpock: https://github.com/cvra/robot-software/blob/master/master-firmware/src/priorities.h ?

master-firmware/src/udp_topic_injector.c
```diff
@@ -0,0 +1,43 @@
...
+
+    struct netconn* conn;
+
+    chThdSetPriority(HIGHPRIO);
+    chRegSetThreadName(__FUNCTION__);
+
+    /* Listens to UDP on port 10'000 */
+    conn = netconn_new(NETCONN_UDP);
+    chDbgAssert(conn != NULL, "Cannot create a connection object");
+    netconn_bind(conn, IPADDR_ANY, 10000);
```
SyrianSpock: port as param?

antoinealb: That would also require watching the port for changes, and reclosing/reopening it if the port changes. Considered out of scope.

master-firmware/src/udp_topic_injector.c
SyrianSpock: it's confusing, why not create it as `HIGHPRIO` below? and it's good point, why`HIGHPRIO`? I'd expect debugging tasks to be `NORMALPRIO` or `LOWPRIO`

antoinealb: I am not sure if it warrants being entered in the priority table. This implementation can only work if this thread is of the higher priority, otherwise messages will be lost. Plus priorities.h is not even used by a lot of people.

master-firmware/src/udp_topic_injector.c
```diff
@@ -0,0 +1,43 @@
...
+        /* Read a datagram */
+        if (netconn_recv(conn, &buf) != ERR_OK) {
+            continue;
+        }
+
+        /* TODO: Better than a static object ? */
+        static uint8_t msg[1024];
+        netbuf_copy(buf, msg, sizeof(msg));
+
+        messagebus_inject_encoded_message(&bus, msg, netbuf_len(buf));
```
SyrianSpock: as if it was meant to work with UDP :stuck_out_tongue:

antoinealb: ? What do you mean ?

SyrianSpock: thatwasajoke

master-firmware/src/udp_topic_injector.c
SyrianSpock: OK

master-firmware/src/ally_position_service.c
```diff
@@ -113,6 +113,6 @@ void ally_position_start(void)
     static THD_WORKING_AREA(receive_wa, 2048);
     parameter_namespace_t* ip_ns = parameter_namespace_find(&global_config, "/ip");
     parameter_string_declare(&ally_ip, ip_ns, "ally_address", ally_ip_buffer, sizeof(ally_ip_buffer));
-    chThdCreateStatic(send_wa, sizeof(send_wa), NORMALPRIO, position_send_thread, NULL);
-    chThdCreateStatic(receive_wa, sizeof(receive_wa), NORMALPRIO, position_receive_thread, NULL);
+    // chThdCreateStatic(send_wa, sizeof(send_wa), NORMALPRIO, position_send_thread, NULL);
+    // chThdCreateStatic(receive_wa, sizeof(receive_wa), NORMALPRIO, position_receive_thread, NULL);
```
SyrianSpock: can we remove this altogether?

master-firmware/src/udp_topic_injector.c
SyrianSpock: > This implementation can only work if this thread is of the higher priority

Sad. We can't do `chThdSetPriority(HIGHESTPRIO());` :crying_cat_face:. Why do you need the higher priority? I'm worried about the control/odometry threads. :thinking: 

> Plus priorities.h is not even used by a lot of people.

Sure, but if it's the way to go, we can make every thread use it. RE-FAC-TO-RING!G

antoinealb: HIGHPRIO is by defintion the highest priority supported by the ChibiOS kernel IIRC. We need the highest priority here to avoid dropping messagebus messages, although that might be acceptable. I don't think the odometry thread will be starved by this more than anything, the position manager is already running at NORMALPRIO, along with everything else.

> Sure, but if it's the way to go, we can make every thread use it. RE-FAC-TO-RING!

Sure, but then we should file a bug and someone should implement it.

SyrianSpock: > HIGHPRIO is by definition the highest priority supported by the ChibiOS kernel IIRC
 
I didn't know that, seems OK.

> We need the highest priority here to avoid dropping messagebus messages, although that might be acceptable.

Not acceptable to drop messages? :thinking: 

> I don't think the odometry thread will be starved by this more than anything, the position manager is already running at NORMALPRIO, along with everything else.

Fair enough, can we verify this when running a hardware test?

> Sure, but then we should file a bug and someone should implement it.

Ask no more: #265

master-firmware/src/ally_position_service.c
SyrianSpock: ?

PR #252, opened 2019-11-06
Actuator board firmware skeleton

antoinealb:
Not tested on any hardware yet, but starts the software project for this board.


antoinealb:
It will most likely not be hardware tested before the merge. That's because I want to start working on this firmware ASAP, to have it ready for testing and debugging once the PCBs arrive. On the other hand I don't want to do a giant 3k LOC pull request in one go, so I believe the best is reviewing that one and merging it without testing.

For now I realized that this new firmware is not covered by our CI, so that's definitely something we should have before merge.

SyrianSpock:
Sounds fair, let's go like this, but how far are we from testing on hardware to validate this?

PR #253, opened 2019-11-14
Honeywell pressure sensor driver for pump board

antoinealb:
Not tested on hardware, as we don't have hardware yet.

**Note:** This shows more commit than there actually is in this PR. It needs to be rebased once #252 lands.


SyrianSpock:
Merge?

PR #256, opened 2019-12-11
Delete 2019 code

antoinealb:
This PR removes the code that was specific to the 2019 robot so that we can all start working on the 2020 project :)

@SyrianSpock: it removes a lot of your code, not too sad to see it go ?


master-firmware/src/strategy.cpp
```diff
@@ -168,53 +105,23 @@ static void wait_for_autoposition_signal(void)
...
-    return res;
-}
-
-void strategy_shutdown_endgame(void)
-{
-    manipulator_gripper_set(BOTH, GRIPPER_OFF);
-}
-
-void strategy_order_play_game(strategy_context_t* ctx, RobotState& state)
+void strategy_order_play_game(RobotState& state, enum strat_color_t color)
```
SyrianSpock: pointers are more in your face IMHO

antoinealb: Not sure I understand the comment here

SyrianSpock: I would pass `RobotState` as a pointer instead of a reference

SyrianSpock: But then it could be null #NoFreeLunch 😭

antoinealb: Yes I think I will keep as non nullable for now and merge it like this.

SyrianSpock: :shipit:

PR #260, opened 2020-01-23
Format Python code on push automatically

antoinealb:


can-io-firmware/send_pwm.py
```diff
@@ -7,23 +7,21 @@
 import uavcan
 import os
 
-DSDL_DIR = os.path.join(os.path.dirname(__file__), '../uavcan_data_types/cvra')
+DSDL_DIR = os.path.join(os.path.dirname(__file__), "../uavcan_data_types/cvra")
```
SyrianSpock: Why double quote when you can single quote?

tools/arm_graph/parser.py
```diff
@@ -7,49 +7,53 @@
 def parse(file):
```
SyrianSpock: Can we delete this file?

tools/config/config_to_c.py
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env python3
```
SyrianSpock: Do we still need this file?

antoinealb: I dont know. That is not in the scope of this PR in my opinion.

SyrianSpock: True, but still wondering

can-io-firmware/send_pwm.py
antoinealb: We just need to pick one and stick to it. So far it is mixed, even in a single file.

tools/config/config_to_c.py
antoinealb: I don't think for that one. Maybe for the parser in the other comment.

PR #261, opened 2020-01-23
Weather vane orientation detection

antoinealb:
* Tested for real.
* Sends result over UDP/Protobuf directly into a messagebus topic
* Needs some more flexible config, but already sweet sweet proof of concept.

Collaboration brought to you by @nuft and I.


computer-vision/README.md
```diff
@@ -0,0 +1,26 @@
+# ArUco Robot Tracking
```
SyrianSpock: So technically not related to this PR

computer-vision/config/iphone.yaml
```diff
@@ -0,0 +1,23 @@
+%YAML:1.0
```
SyrianSpock: So works on an iPhone for now?

computer-vision/robot_tracking.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+enum WeathiervaneDirection {
+    NORTH = 0,
+    SOUTH = 1
+};
+
+WeathiervaneDirection weathervane_direction(cv::Vec3d rvec, cv::Vec3d tvec)
+{
+    cv::Mat R;
+    cv::Rodrigues(rvec, R);
```
SyrianSpock: aaaaah Rodrigues convention :volcano:

computer-vision/robot_tracking.cpp
```diff
@@ -0,0 +1,167 @@
...
+    for (size_t i = 0; i < ids.size(); i++) {
+        if (ids[i] == ARUCO_WEATHERVANE) {
+            dir = weathervane_direction(rvecs[i], tvecs[i]);
+            break;
+        }
+    }
+
+    switch (dir) {
+        case NORTH:
+            printf("NORTH\n");
```
SyrianSpock: printf, really?
```suggestion
            std::cout << "NORTH"  << std::endl;
```

computer-vision/robot_tracking.cpp
```diff
@@ -0,0 +1,167 @@
...
+    result.set_weather_vane_orientation(ComputerVisionResult::NORTH);
+    return result;
+}
+
+int main(int argc, char* argv[])
+{
+    logging_init();
+    if (argc < 2) {
+        ERROR("usage: %s config.yaml image.jpg\n", argv[0]);
+        return -1;
```
SyrianSpock:
```suggestion
        return EXIT_FAILURE;
```
or
```suggestion
        throw("usage: %s config.yaml <image.jpg>\n", argv[0]);
```

computer-vision/tools/grab_image.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
+import cv2
+
+cap = cv2.VideoCapture(0)
+
+n = 0
+while(n < 10):
```
SyrianSpock: no main?

computer-vision/README.md
antoinealb: Oh oops, forgot to change the content of the README.

computer-vision/config/iphone.yaml
antoinealb: we can use pictures taken by an iPhone but it does not run on the phone.

computer-vision/robot_tracking.cpp
antoinealb: This should also be removed. And rather than std::cout i would rather have INFO macro

computer-vision/robot_tracking.cpp
antoinealb: Yes, need to cleanup the code. This is more of a proof of concept

computer-vision/tools/grab_image.py
antoinealb: Same, consider this proof of concept. The python scripts were not really cleaned up yet.

computer-vision/config/iphone.yaml
nuft: It was the best camera available for testing

computer-vision/robot_tracking.cpp
antoinealb: What is rodrigues convention ?

SyrianSpock: no, just thesis flashbacks #vietnam

SyrianSpock: the day we lost @antoinealb https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula

computer-vision/robot_tracking.cpp
nuft: This was built upon an opencv example. Focus was on functionality not style. We'll refactor later or in another PR.

SyrianSpock: I'd still refactor here, it's OK, I just don't care it's copied from an example, I'd like it cleaned up before merge

computer-vision/tools/grab_image.py
SyrianSpock: mkay, also no function to be reused, so no reason to import the file

computer-vision/robot_tracking.cpp
```diff
@@ -0,0 +1,171 @@
...
+        destination_port = (int)fs["destination_port"];
+    }
+};
+
+enum WeathiervaneDirection {
+    NORTH = 0,
+    SOUTH = 1
+};
+
+WeathiervaneDirection weathervane_direction(cv::Vec3d rvec, cv::Vec3d tvec)
```
SyrianSpock:
```suggestion
WeathervaneDirection weathervane_direction(cv::Vec3d rvec, cv::Vec3d tvec)
```
typo?

master-firmware/protobuf/computer_vision.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
+
+import "nanopb.proto";
+
+message ComputerVisionResult {
```
SyrianSpock: `std::optional`?

computer-vision/README.md
```diff
@@ -0,0 +1,21 @@
+# CVRA 2019 Computer vision
+
+
+Requirements:
```
SyrianSpock:
```suggestion
## Requirements
```

computer-vision/README.md
```diff
@@ -0,0 +1,21 @@
...
+## Camera calibration
+
+Follow the OpenCV tutorial: [Camera calibration With OpenCV](https://docs.opencv.org/2.4/doc/tutorials/calib3d/camera_calibration/camera_calibration.html).
+You may also read [Camera Calibration and 3D Reconstruction](https://docs.opencv.org/2.4/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html).
+
+Take some picture with the script `tools/grap_image.py` and then run `tools/calibrate.py` to get the camera matrix and distortion coefficient vector.
+
+Note: Fisheye camera need different procedure using [fisheye::calibrate](https://docs.opencv.org/2.4/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html?highlight=calibratecamera#fisheye-calibrate)
+
+## ArUco tracking
```
SyrianSpock: Add a section about how to run it?

master-firmware/protobuf/computer_vision.proto
antoinealb: I am not sure what you suggest here. This is a proto file, not c++.

computer-vision/README.md
antoinealb: We will. But it is not finished yet, so for now there is no "how to run".

computer-vision/robot_tracking.cpp
antoinealb: Fixed by removing the intermediate enum in favor of the protobuf-declared one.

master-firmware/protobuf/computer_vision.proto
SyrianSpock: Someday we'll get optional in protobuf? A man can dream

computer-vision/README.md
SyrianSpock: Not even a "how to dev"?

computer-vision/robot_tracking.cpp
```diff
@@ -0,0 +1,157 @@
...
+    cv::Rodrigues(rvec, R);
+    cv::Vec3d e2 = {0, 1, 0};
+    cv::Vec3d v = cv::Mat(R * e2);
+
+    // check if facing down
+    if (v[1] > 0) {
+        return ComputerVisionResult::SOUTH;
+    } else {
+        return ComputerVisionResult::NORTH;
+    }
```
SyrianSpock: No invalid result possible?

computer-vision/README.md
SyrianSpock: No how to dev instructions? How do you test this code?

SyrianSpock:
I don't get the error on CI,operator* should perform matrix multiplication 🤔

nuft:
> I don't get the error on CI, operator* should perform matrix multiplication

I had the same compilation error on the RasPi. Need to investigate...

antoinealb:
The versions of OpenCV are different between macOS (4.2.0) and Ubuntu Bionic (3.2.0), that could be a source of headaches.

Now I am wondering if we should perhaps build OpenCV as part of the build ?

nuft:
We should build against the same version. I'm for the more recent one.
Also I would specify the version requirement in the CMakeLists.txt so that it fails if not satisfied.
Though, I'm not sure if automatically pulling the source and building it here is the best way to do it.

PR #263, opened 2020-01-25
[WIP] Add table objects

Lu-ni:
Not to be merged yet,

I'm opening this to ask what we need more in table.h.
Also, how does the mirroring function work ? because the color of the gobelet are reversed depending on the side of the table.

I'll use this PR for map.c too

### DIFF
--- a/master-firmware/package.yml
+++ b/master-firmware/package.yml
@@ -76,6 +76,8 @@ source:
     - src/base/base_helpers.c
     - src/strategy/state.cpp
     - src/strategy/score.cpp
+    - src/strategy/actions_goap.cpp
+    - src/strategy/goals.cpp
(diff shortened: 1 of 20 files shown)


SyrianSpock:
> I'm opening this to ask what we need more in table.h.

I think it's useless to guess, we need to write the game strategy to figure it out

> Also, how does the mirroring function work? because the color of the goblet is reversed depending on the side of the table.

which macro? doesn't make sense to mirror colors if it's not mirrored.

SyrianSpock:
Mergeable soon?

Lu-ni:
> which macro? doesn't make sense to mirror colors if it's not mirrored.
Depending on the side, the colors are reversed

PR #266, opened 2020-03-11
Actuator board: Basic readout functionality

antoinealb:
This pull requests implements some basic readout of the board's analog input (for testing), as well as the status of the pressure sensors.

It currently reports if the pressure sensor has any faults through the NodeStatus message. I guess next step would be to actually publish a pressure, but for that I would need a physical board.

not tested, as hardware has not arrived yet.

### DIFF
--- a/actuator-firmware/Makefile
+++ b/actuator-firmware/Makefile
@@ -78,7 +78,7 @@ include $(CHIBIOS)/os/common/ports/ARMCMx/compilers/GCC/mk/port_v7m.mk
 # Define linker script file here
 
 ifeq ($(USE_BOOTLOADER), )
-	USE_BOOTLOADER = no
+	USE_BOOTLOADER = yes
 endif
 
 ifeq ($(USE_BOOTLOADER), yes)
--- a/actuator-firmware/src/analog_input.c
+++ b/actuator-firmware/src/analog_input.c
@@ -1,0 +1,34 @@
+#include <ch.h>
+#include <hal.h>
+
+#include "hal_adc.h"
+#include "hal_adc_lld.h"
+
+/* We have an external attenuation of 0.5 before reaching the analog input */
+static const float ADC_GAIN = 2 * 3.3 / 4096;
+
+static const ADCConversionGroup group = {
+    .circular = true,
+    .num_channels = 2,
+    .end_cb = NULL,
+    .error_cb = NULL,
+    .cfgr = 0,
+    /* Sample every channel for 600 ADC clock cycles (SMP=5) .*/
+    .smpr = {ADC_SMPR1_SMP_AN0(5) | ADC_SMPR1_SMP_AN1(5), 0},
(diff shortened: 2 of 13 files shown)


actuator-firmware/Makefile
```diff
@@ -78,7 +78,7 @@ include $(CHIBIOS)/os/common/ports/ARMCMx/compilers/GCC/mk/port_v7m.mk
 # Define linker script file here
 
 ifeq ($(USE_BOOTLOADER), )
-	USE_BOOTLOADER = no
+	USE_BOOTLOADER = yes
```
SyrianSpock: :+1:

actuator-firmware/src/board.c
```diff
@@ -112,3 +114,8 @@ void board_reset_pressure_sensors(void)
     chThdSleepMilliseconds(1);
     palSetPad(GPIOA, GPIOA_PRESSURE_RST);
 }
+
+int board_digital_input_read(void)
```
SyrianSpock: :+1:

uavcan_data_types/cvra/actuator/20151.Command.uavcan
```diff
@@ -0,0 +1,9 @@
+#
+# Command for an actuator board
+#
+
+# UAVCAN node ID for unicast addressing
+uint7 node_id
+
+float16[2] pumps     # PWM in range 0-1
+bool[2]    solenoids # [V]
```
nuft: This is a bool with no unit []

uavcan_data_types/cvra/actuator/20150.Feedback.uavcan
```diff
@@ -0,0 +1,7 @@
+#
+# Feedback coming from actuator-firmware nodes
+#
+
+float16[2] pressures       # [Pascal]
```
nuft: In other message definitions we use singular for fields, even in case of an array.
Personally I prefer singular because eg. `feedback.analog_input[1]` looks nicer.

actuator-firmware/src/analog_input.c
```diff
@@ -0,0 +1,34 @@
+#include <ch.h>
+#include <hal.h>
+
+#include "hal_adc.h"
+#include "hal_adc_lld.h"
+
+/* We have an external attenuation of 0.5 before reaching the analog input */
+static const float ADC_GAIN = 2 * 3.3 / 4096;
+
+static const ADCConversionGroup group = {
```
nuft: This looks correct. I didn't go through the datasheet though.

antoinealb: ADC configs are very error-prone, so I expect some mistakes to show up in there anyways :smile:

uavcan_data_types/cvra/actuator/20150.Feedback.uavcan
antoinealb: done

PR #267, opened 2020-03-22
Implement an API to log a message every N call

antoinealb:
This new API allows easier debugging of very frequent loops (such as
control loops) without spamming the terminal.

Note that there is no ERROR_N message, as I don't think we should ever
silently drop errors.

### DIFF
--- a/lib/error/error.h
+++ b/lib/error/error.h
@@ -99,6 +99,19 @@ const char* error_severity_get_name(uint8_t severity);
         }                                                           \
     } while (0)
 
+/** Call this macro to only log one out of N occurence. This is useful in tight call loops. */
+#define WARNING_EVERY_N(n, text, ...)                               \
+    do {                                                            \
+        static int _cnt = 0;                                        \
+        if ((_cnt++ % n) == 0 && g_error_fct.warning) {             \
+            struct error e = error_generate(ERROR_SEVERITY_WARNING, \
+                                            (text),                 \
+                                            (__FILE__),             \
+                                            __LINE__);              \
+            g_error_fct.warning(&e, ##__VA_ARGS__);                 \
+        }                                                           \
+    } while (0)
+
 /** Call this macro to log NOTICE events */
 #define NOTICE(text, ...)                                          \
     do {                                                           \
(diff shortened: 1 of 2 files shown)


lib/error/error.h
```diff
@@ -99,6 +99,19 @@ const char* error_severity_get_name(uint8_t severity);
         }                                                           \
     } while (0)
 
+/** Call this macro to only log one out of N occurence. This is useful in tight call loops. */
+#define WARNING_N(n, text, ...)                                     \
```
SyrianSpock: no templates? :slightly_smiling_face: `WARNING<N>` with a default to 1 would be nicer

SyrianSpock:
```suggestion
template<int n = 1> void WARNING(const std::string& text, ...) {
        static int _cnt = 0;                                       
        if ((_cnt++ % n) == 0 && g_error_fct.warning) {             
            struct error e = error_generate(ERROR_SEVERITY_WARNING, 
                                            (text),                 
                                            (__FILE__),             
                                            __LINE__);              
            g_error_fct.warning(&e, ##__VA_ARGS__);                 
        }                                                           
}
```

lib/error/tests/error.cpp
```diff
@@ -69,6 +69,57 @@ TEST(ErrorLogging, GeneratesDebug)
...
+    mock().expectOneCall("error").withStringParameter("text", "bar").ignoreOtherParameters();
+
+    for (int i = 0; i < 20; i++) {
+        // Log every 10 loop iteration
+        NOTICE_N(10, "foo");
+        NOTICE_N(10, "bar");
+    }
+}
+
+TEST(ErrorLogging, GeneratesDebugEveryN)
```
SyrianSpock:
```suggestion
TEST(ErrorLogging, GeneratesDebugEvery10)
```

lib/error/tests/error.cpp
```diff
@@ -69,6 +69,57 @@ TEST(ErrorLogging, GeneratesDebug)
...
+    mock().expectOneCall("error").withStringParameter("text", "bar").ignoreOtherParameters();
+
+    for (int i = 0; i < 20; i++) {
+        // Log every 10 loop iteration
+        WARNING_N(10, "foo");
+        WARNING_N(10, "bar");
+    }
+}
+
+TEST(ErrorLogging, GeneratesNoticeEveryN)
```
SyrianSpock:
```suggestion
TEST(ErrorLogging, GeneratesNoticeEvery10)
```

lib/error/tests/error.cpp
```diff
@@ -69,6 +69,57 @@ TEST(ErrorLogging, GeneratesDebug)
     DEBUG("foo");
 }
 
+TEST(ErrorLogging, GeneratesWarningEveryN)
```
SyrianSpock:
```suggestion
TEST(ErrorLogging, GeneratesWarningEvery10)
```

lib/error/error.h
antoinealb: This would not work for several reasons:

1. This API is used by pure C callers. While we discuss from time to time migrating to complete C++, this is not something that happened yet.
2. We need preprocessor expansion for __LINE__ and __FILE__, not sure how that would play with templates.
3. The API is not very clean: What does `WARNING<4>("foobar")` mean?

Also, wouldn't that lead to a much bigger code size ?

SyrianSpock: `WARNING_N` what does it mean? both are arbitrary IMO but points 1-2 remain valid same for code size

antoinealb: Okay, what I meant is that templates here do not provide additional cleanliness compared to what can be achieved in good old  C.

antoinealb: Renamed to `WARNING_EVERY_N` to try to make it more readable.

antoinealb:
Test renamed.

PR #268, opened 2020-03-22
Implement servomotors handling for actuator board

antoinealb:
This pull request implements control of the servomotor output that are on the actuator board. They can be controlled through UAVCAN and support having a smooth acceleration.

Tested on hardware (with a frequency meter and voltmeter as I don't have any RC servos).

Note that this PR depends on #267 as it uses the new DEBUG_N calls.

### DIFF
--- a/actuator-firmware/src/pwm.h
+++ b/actuator-firmware/src/pwm.h
@@ -8,17 +8,32 @@ extern "C" {
 #include <ch.h>
 #include <hal.h>
 
-enum pwm_channel {
-    PWM_CHANNEL_0, // Timer 16 channel 0
-    PWM_CHANNEL_1, // Timer 17 channel 0
-    PWM_CHANNEL_2, // Timer 1 channel 1
-    PWM_CHANNEL_3, // Timer 1 channel 2
-};
+/// Mask to select channel number, 1 indexed
+#define PWM_CHANNEL(x) (1 << (x - 1))
 
-void pwm_init(uint32_t frequency, uint32_t period);
+/** Low level init of PWM channels for a timer.
+ *
+ * @param [in] tim Address of the timer to use, example STM32_TIM1.
+ * @param [in] channel_mask Which channels to enable, OR'd together. You can
+ * use PWM_CHANNEL(n) to enable channel N.
+ * @param [in] clock_frequency The frequency of the clock running this
+ * particular timer peripheral. You can use macros such as STM32_TIMCLK2 to
+ * compute it from the RCC configuration.
+ * @param [in] timer_frequency The frequency at which the timer should count.
+ * @param [in] timer_period The timer period.
+ */
+void pwm_setup_channels(stm32_tim_t* tim,
+                        uint8_t channel_mask,
+                        uint32_t clock_frequency,
+                        uint32_t timer_frequency,
+                        uint32_t timer_period);
 
-/* Set given pulse width in clock ticks to the corresponding PWM channel */
-void pwm_set_pulsewidth(enum pwm_channel channel, uint32_t pulsewidth);
+/** Sets the pulse width for that particular timer and channel.
+ *
+ * Note that the channel number is 0-indexed here, while in the datasheet it is
+ * 1-indexed.
+ */
+void pwm_set_pulsewidth(stm32_tim_t* tim, uint8_t channel, uint32_t width);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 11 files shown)


actuator-firmware/src/pwm.h
```diff
@@ -8,17 +8,32 @@ extern "C" {
...
+                        uint32_t timer_period);
 
-/* Set given pulse width in clock ticks to the corresponding PWM channel */
-void pwm_set_pulsewidth(enum pwm_channel channel, uint32_t pulsewidth);
+/** Sets the pulse width for that particular timer and channel.
+ *
+ * Note that the channel number is 0-indexed here, while in the datasheet it is
+ * 1-indexed.
+ */
+void pwm_set_pulsewidth(stm32_tim_t* tim, uint8_t channel, uint32_t width);
```
SyrianSpock: Maybe we want also an API with on/off and configurable pwm for those

antoinealb: What would you want to configure ?

SyrianSpock: on/off pulsewidth

antoinealb: Well the on pulse width is specified from the duty cycle, so once per channel (hence this function). On the other hand, the cycle duration can only be specified on a per-timer basis (i.e. for all channels at once), and this is why it is provided as a parameter to `pwm_setup_channels`.

actuator-firmware/src/uavcan/Command_handler.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <cvra/actuator/Command.hpp>
+
+#include "uavcan/build_config.hpp"
+
+int Command_handler_start(uavcan::INode& node)
+{
+    int ret;
+    static uavcan::Subscriber<cvra::actuator::Command> sub(node);
+    ret = sub.start(
```
nuft: It's good practice to initialize local variables at declaration. Or in this case, return directly.
And always use const when possible.

actuator-firmware/src/servo.c
```diff
@@ -100,6 +95,9 @@ THD_FUNCTION(servo_thd, arg)
 
 void servo_start(void)
 {
-    static THD_WORKING_AREA(servo_thd_wa, 256);
+    static THD_WORKING_AREA(servo_thd_wa, 512);
+
+    servo_init();
```
nuft: `servo_init()` -> `timer_init()`

actuator-firmware/src/servo.c
```diff
@@ -100,6 +95,9 @@ THD_FUNCTION(servo_thd, arg)
 
 void servo_start(void)
 {
-    static THD_WORKING_AREA(servo_thd_wa, 256);
+    static THD_WORKING_AREA(servo_thd_wa, 512);
```
nuft: We should monitor actual stack usage and not just randomly double.
This is ok for development but we might run into memory shortage on this constrained MCU.

actuator-firmware/src/uavcan/Command_handler.cpp
```diff
@@ -0,0 +1,35 @@
...
+                return;
+            }
+            /* TODO: Implement a watchdog to stop the movements if no setpoints come from master */
+            /* TODO: Handle commands other than servomotors. */
+
+            UAVCAN_ASSERT(SERVO_COUNT == msg.servo_trajectories.size());
+
+            for (int i = 0; i < SERVO_COUNT; i++) {
+                servo_set(i, msg.servo_trajectories[i].position, msg.servo_trajectories[i].velocity, msg.servo_trajectories[i].acceleration);
+                DEBUG_N(10, "received setpoint %03d for servo %d", static_cast<int>(msg.servo_trajectories[i].position * 1000), i);
```
nuft: DEBUG_N -> DEBUG_EVERY_N

actuator-firmware/src/uavcan/Command_handler.cpp
antoinealb: done.

actuator-firmware/src/servo.c
antoinealb: servo_init does more than timer_init, and is a separate function.

actuator-firmware/src/servo.c
antoinealb: +1, not even sure that stack size is an issue here.

actuator-firmware/src/uavcan/Command_handler.cpp
antoinealb: done

PR #269, opened 2020-03-24
Control pumps and solenoids from UAVCAN

antoinealb:
Tested on hardware, works.

Also added a GUI interface in studio to control the servomotors and pumps easily.

### DIFF
--- a/actuator-firmware/package.yml
+++ b/actuator-firmware/package.yml
@@ -13,18 +13,19 @@ depends:
   - quadramp
 
 target.arm:
-  - src/main.c
-  - src/debug.c
+  - ../lib/can-bootloader/config.c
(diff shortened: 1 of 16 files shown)


nuft:
Nice GUI, works and commands are received.
But when connecting the solenoid & pump nothing moves.
@antoinealb have you also measured if PWM output is correct?

antoinealb:
Yes I tested it, and also the power part using a lab power supply feeding on vbat. That's weird that it does not work for you

nuft:
Ok, this must be a hardware issue then.

nuft:
Issues came from faulty hardware. Tested for both Pump1/2 and solenoid 1/2.
Can't test servo atm but this is low risk.

PR #272, opened 2020-05-19
[WIP] Wheel Encoder Firmware

nuft:


uavcan_data_types/cvra/odometry/20160.WheelEncoder.uavcan
```diff
@@ -0,0 +1,6 @@
+#
+# Differential base Encoder Position
+#
+
+uint16 left_encoder_raw
```
antoinealb, line 5: I think it would make more sense to have them as 32 bit values, 16 bit will overflow quite early. We can also choose to handle the overflow, but it is just added complexity.

nuft: Depends if the timers support 32bit. Also, we're sending with 100Hz, I don't think this will overflow between two messages.

antoinealb: Looking at the `motor-control-firmware` code, it seems that the QEI support 32 bits.

I don't think we should base our API definitions on the hardware. I would rather have the overflow handled locally, and the non-overflowing value sent over UAVCAN, in order to have the board always send a value that make sense.

antoinealb: So yesterday we discussed this. The idea that we reached was the following: loosing one CAN frame should not be an issue. So our implementation publishes a processed `int32` that contains the already-integrated value.

nuft:
> Have you thought about organizing the code in a similar way to what we did with the proximity beacon ?

There is very little code shared between wheel-encoder and motor-control firmwares.
If we really plan to go forward with a dedicated board I would like to keep it separate.

Now that the master firmware modifications are added in this branch I wonder if it is meant to be merged into master?

antoinealb:
> Now that the master firmware modifications are added in this branch I wonder if it is meant to be merged into master?

We can always remove the changes to master before merging, they live in their own little commits. I just did not want to have them only on my PC :)

nuft:
Ok for me. I believe it is not much effort to replicate the functionality on a motor-control-board. So once this is ready I will revert and merge.

antoinealb (approved):

PR #282, opened 2020-11-01
Light up a LED if emergency button is pressed

antoinealb:
If the emergency shutdown button is depressed, then the battery voltage
will be cut to the motor control boards, but they will still be powered
by the CAN bus.

This patch reports the battery powerdown from the motor control board as
a custom status code. On the master firmware we simply store the status
of the boards and report them through a LED.

Needs to be tested on hardware.


antoinealb:
tested on hw, merging

PR #287, opened 2020-11-18
Update ARM GCC to latest version

antoinealb:
Mostly as currently the actuator board fails to build on the CI but works on my machine, so I want to see if upgrading the compiler version changes something.


antoinealb:
Closing as the CI is now correctly working and migrated.

PR #300, opened 2021-09-01
Implement software SPI workaround for pressure sensor

antoinealb:
@nuft identified that for some reason there is a conflict between the
SPI driver and the servo motor PWN; turning on the servo means the SPI
hardware starts misbehaving immediately. This is likely not a PCB issue,
as the problem starts as soon as the peripheral is enabled, even if no
signal is coming out of the chip.

Two workarounds were possible: either using a software PWM or a software
SPI. Servomotors require rather precise PWM signals, as the use pulse
width to determine which position they should have. Therefore, I am more
incline toward using a software SPI, which despite being slow, requires
no accurate timings.

This commit implements a basic software emulation of SPI, with just the
modes required for communication with the pressure sensor. I could not
test it on hardware, but I have written unit tests for it. Support for
calibrating the data rate is present, but I don't have the required
instruments to do it available.

Note that since this SPI driver is based on busy waiting, the calling
thread will be using CPU, so some thread priorities might need to be
adjusted.


nuft:
Looks good to me. I’ll test it on hardware asap.

nuft (approved):
Works great!

PR #301, opened 2021-09-02
Fix pressure sensor SPI issues

nuft:
Includes fix from PR #300 which I tested on hardware - it works great 🥳 

This also fixes an issue where one sensor gets stuck if the other sensor is addressed on the same SPI bus. So before each read both sensors are reset.

A useful `print_stack_info()` routine was added to properly choose the thread stack size. It seems like we actually have plenty of RAM:
```
stackptr  stacktop  stklimit  free   name
200006AC  200004C8  20000444    132  main
20000BEC  20000BEC  20000BE4      8  idle
20001064  20001064  20000FEC    120  heartbeat
20001394  20001304  200011EC    280  servo
20001B3C  200018E8  2000160C    732  uavcan
```


actuator-firmware/src/debug.c
```diff
@@ -76,3 +76,38 @@ void panic_handler(void)
...
+
+void print_stack_info(void)
+{
+#if (CH_DBG_FILL_THREADS != TRUE) || (CH_CFG_USE_REGISTRY != TRUE) || (CH_DBG_ENABLE_STACK_CHECK != TRUE)
+#error "Requires: CH_DBG_FILL_THREADS CH_CFG_USE_REGISTRY CH_DBG_ENABLE_STACK_CHECK"
+#endif
+
+    BaseSequentialStream* chp = OUTPUT_STREAM;
+
+    const uint32_t STACK_FILL = 0x55555555;
```
antoinealb, line 88: Consider using CH_DBG_STACK_FILL_VALUE rather than a "magic value". Will required you to move to checking byte-by-byte rather than word-by-word.

actuator-firmware/src/uavcan/feedback_publisher.cpp
```diff
@@ -8,27 +8,43 @@
...
+    palSetPad(GPIOA, GPIOA_PRESSURE_RST);
+    chThdSleepMilliseconds(3);
+
+    mpr_start_measurement(&pressure_sensors[sensor]);
+    chThdSleepMilliseconds(5); // Datasheet says wait at least 5m for conversion  (Datasheet p.15)
+
+    // error condition
+    const uint8_t status = mpr_read_status(&pressure_sensors[sensor]);
+    if (mpr_status_is_error(status) || mpr_status_is_busy(status)) {
+        WARNING("Pressure sensor %d status error: 0x%x", sensor, status);
```
antoinealb, line 26: If that happens, can we somehow reflect this in the UAVCAN node status?

antoinealb (approved):
Please rebase on master before merging.

actuator-firmware/src/debug.c
nuft: Ok done

actuator-firmware/src/uavcan/feedback_publisher.cpp
nuft: done

antoinealb (approved):

PR #302, opened 2021-11-26
Prevent page out for master-firmware memory.

antoinealb:
Paging out means storing part of the software's memory to disk and
occurs when we use more memory than what is available on the system.
This can cause high latency in some memory access, which is not possible
in realtime operations. This can be prevented by "locking" some memory,
which instructs the kernel to never page out memory.

This patch adds memory locking to the master firmware. As locking out
memory requires special capabilities or root, the call to memory locking
is put behind a flag that is disabled by default, and enabled on the
robot. On the robot we must also increase the maximum mlock size, as by
default it will only be a few kilobytes.

Ressources:

* https://lwn.net/Articles/876288/
* https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_for_real_time/7/html/reference_guide/using_mlock_to_avoid_page_io

### DIFF
--- a/master-firmware/packaging/cvra-master.service
+++ b/master-firmware/packaging/cvra-master.service
@@ -6,6 +6,7 @@ Type=simple
 User=root
 # TODO(antoinealb): What do we want to do with the onboard LED?
 ExecStart=/bin/xinit /bin/cvra-master --enable_gui --can_iface="can0" \
+    --lock_memory \
     --led_comm_path=/sys/class/leds/onboard \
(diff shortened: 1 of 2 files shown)


antoinealb:
I still need to test it on hardware, I only checked that it builds.

nuft:
I wonder how often paging out would happen.But it's a good idea to do this.

Another trick I heard of is CPU pinning for the real-time process to guarantee some compute resource.

nuft (approved):

